PR #3, opened 2016-10-24
fix small autocomplete typo

DorianScholz:
This looks like a great project.
I just had a small error when compiling.
This std::__cxx11::string probably just slipped in through auto-complete, since it is used nowhere else.
Removing it made it compile for me on Ubuntu 14.04 with gcc 4.8.4.


facontidavide:
Thank you very much for reporting.

PR #4, opened 2016-10-24
Main Menu fix

DorianScholz:
For some reason the main window's menu bar does not show up on Ubuntu 14.04 using Gnome Flashback Session.
Adding a simple show() call makes it work...


facontidavide:
Thanks for the fix

PR #6, opened 2016-10-25
Small fixes for building with catkin_tools

kartikmohta:
Just a few fixes for issues when building with `catkin_tools` instead of `catkin_make`.


facontidavide:
Ok. Honestly this commit create some minor problems when I try to compile PlotJuggler WITHOUT catkin, but it is still better to support catkin_tools, considering the current user target.

kartikmohta:
What were the issues when building without catkin?

PR #11, opened 2016-12-14
Add plot grid + small updates to legend styling

kartikmohta:
I accept that styling is a very subjective thing, so just take this as a suggestion and feel free to ignore if you don't like it 🙂


facontidavide:
Hi @kartikmohta ,

I appreciate the PR, but I guess that the "right way to do this styling is with a proper "Preferences" panel where the user can set the look and feel and some other options.

This is something that I wanted to do, but I have been postponing it too long.
I will try to find some time to work on it the next week; right now, I prefer to keep this PR in "stand-by".

Thanks for contributing

kartikmohta:
That is perfect.

facontidavide:
I decided to add a button to show-hide the grid as an alternative to this PR. check commit  0eb7b6f

kartikmohta:
Even better! Thanks 🙂

PR #16, opened 2017-01-08
follow ros convention for build destination folders

mikaelarguedas:
This PR uses `${CATKIN_PACKAGE_LIB_DESTINATION}` and `${CATKIN_PACKAGE_BIN_DESTINATION}` variables to allow the build products to be found by ros tools (e.g. `rosrun`)
This will also allow better support of isolated/non-isolated builds.


mikaelarguedas:
My bad, I got it working without the need for this change. I must have missed something in my catkin config when I tried it the first time. Feel free to close this PR if you don't see any value in using the catkin cmake variables

facontidavide:
The only thing I really care about is to store the plugins in the same folder as the executable. Both versions should work just fine. I will merge. Thanks

PR #19, opened 2017-01-14
Plugins need to be located with the main executable

kartikmohta:
Was changed in 43f6781fc829ef5383e1e050980d964bbafbd139


facontidavide:
Thumb up!

PR #29, opened 2017-03-16
Prevent compiler warning if compiling under ROS

Timple:
Adding STATUS to the message will still show it with a manual make but will not be seen by the catkin build system as a warning.


facontidavide:
thanks

PR #39, opened 2017-04-23
Fix image link in tracker.md

eduherminio:
pj-tracker.png was removed in last documentation upgrade, so I guess PlotJuggler_tracker.png wants to be linked instead (at least for the moment).


facontidavide:
:+1:

PR #51, opened 2017-08-24
Potential fix for precision issue when adding time_offset

kartikmohta:
Adding time_offset (which can be a very large number) to the left and right time end points of a data series, we can have precision issues causing errors of the type "invalid X0/X1 range in PlotWidget::maximumRangeY". To prevent this, we shift the left and right values slightly after adding the time_offset so that they are still in the expected range.


kartikmohta:
[Example bag](https://github.com/facontidavide/PlotJuggler/files/1247496/test.zip) which causes the problem when plotting any of the `acceleration` fields.

facontidavide:
thanks

PR #58, opened 2017-10-21
Don't add -g to CMAKE_CXX_FLAGS by default

kartikmohta:
Leads to larger binaries even when building in `Release` mode.


facontidavide:
As you know, I use __backward-cpp__ library to get the call stack when the application crash.
I believe that this might be useful when people downloading a binary of Plotjuggler send error reports.

I am thinking in particular to the people in the ROS community.

The -g option is enabled even in Release to get a more informative message.

kartikmohta:
Yes, that seems to be worth keeping then 🙂

PR #62, opened 2017-11-20
Fixes #61 -  error with non monotonic time

### LINKED ISSUE #61: Loading valid .csv results in "Selected time in notstrictly  monotonic" error
The first time I load a csv file to PlotJuggler there is a pop-up asking to select the time axis. The file is loaded correctly. When I want to load another file with same data structure, there is "Selected time in notstrictly  monotonic" error. 

The only way I can get data to be loaded correctly then, is to change the name of time axis in the csv.

Example files (extensions changed to txt due to Github restrictions):
[test.txt](https://github.com/facontidavide/PlotJuggler/files/1487087/test.txt)
[test2.txt](https://github.com/facontidavide/PlotJuggler/files/1487088/test2.txt)

mactro:
Vector size was already incremented, so time_index was pointing to `correct_index + 1` column.

### DIFF
--- a/plugins/DataLoadCSV/dataload_csv.cpp
+++ b/plugins/DataLoadCSV/dataload_csv.cpp
@@ -169,7 +169,7 @@ PlotDataMap DataLoadCSV::readDataFromFile(const QString &file_name, bool use_pre
             {
                 if( _default_time_axis == qname )
                 {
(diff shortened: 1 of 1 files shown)


facontidavide:
the fix was slightly more complicated than that for general cases. thanks though.

PR #64, opened 2018-01-05
A small plugin creating a websocket server

PhilGauthier:
Hello,
I wrote a small simple plugin based on your sample.
It allow any third party application to connect to Plotjuggler using websocket and push their values.
I though it could be usefull to others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,7 @@ add_subdirectory( plotter_gui )
 
 add_subdirectory( plugins/DataLoadCSV )
 add_subdirectory( plugins/DataStreamSample )
+add_subdirectory( plugins/DataStreamServer )
 
 if(COMPILING_WITH_CATKIN)
     add_subdirectory( plugins/ROS )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,14 @@ add_subdirectory( plotter_gui )
 
 add_subdirectory( plugins/DataLoadCSV )
 add_subdirectory( plugins/DataStreamSample )
-add_subdirectory( plugins/DataStreamServer )
+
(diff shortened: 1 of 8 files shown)


PhilGauthier:
Erf the continuous integration test fails, because it uses Qt5.2 (from Ubuntu trusty repository).
Websockets are available only from Qt5.3, I don't know if an upgrade of the target environment is planned or an option.

JimmyDaSilva:
👍

facontidavide:
Hi @PhilGauthier ,

congrtulation, you are the first contributor that submit a brand new plugin.
I don't have much time to review it right now, but unfortunately I can not change the minimum requirements of PlotJuggler in terms of Qt version, because I must guarantee that it will compile on trusty.

We should add to the CMakeFiles.txt a way to compile your plugin only if Qt 5.3 or higher was detected.

PhilGauthier:
Hi @facontidavide ,
Thanks for the feedback, yes that's probably the easiest way to go for know.
I added the QT version check to the main CMakeFiles.txt as you proposed.
Cheers!

efernandez:
After this PR was merged I'm having trouble to build PlotJuggler because it can't find `Qt5websockets` in Ubuntu 16.04.

Which debians should be installed to build it now?

I can create an issue if you prefer, but I thought I could start here, since this PR introduced this issue, and all the parts are already here.

PhilGauthier:
@efernandez 👍 
My bad, I was on Windows and made some case errors :
https://github.com/PhilGauthier/PlotJuggler/commit/88aecc5d63a4424fb4a1e80de3d68d9fb43215ba

efernandez:
Oh, I see. Could you create a PR, please?

I confirm it works fixing the case errors.

PR #86, opened 2018-05-31
Reorder header files to fix conflicts with boost and QT

efernandez:


facontidavide:
Hi,

thank you for the PR. i will be happy to merge it, but I am just curious: which error is it and on which platform will the conflict take place?

efernandez:
Thanks for merging @facontidavide .

I recently got a new laptop and I can't tell you exactly the system configuration I had, but it should be similar (likely the same) to this:

- Ubuntu 16.04 xenial
- Boost 1.58.0
- QT 5.5.1
- ROS `melodic-devel`

PR #91, opened 2018-06-27
Devel cch

CorentinChauvin:
Hi,

Here is the pull request we spoke about earlier by email.
What is added for reminder :
- Visualise mean and standard deviation of a signal (new button at the bottom of the gui, icon to be determined and right-click/transforms)
- Visualise FFT of a signal (right-click/transform)

Cheers,
Thomas & Corentin

### DIFF
--- a/plotter_gui/CMakeLists.txt
+++ b/plotter_gui/CMakeLists.txt
@@ -33,6 +33,8 @@ SET( PLOTTER_SRC
     filterablelistwidget.cpp
     tabbedplotwidget.cpp
     subwindow.cpp
+    fft.cpp
+    curvestatistics.cpp
 
     mainwindow.h
     realslider.h
--- a/plotter_gui/curvestatistics.cpp
+++ b/plotter_gui/curvestatistics.cpp
@@ -1,6 +1,6 @@
 #include "curvestatistics.h"
 
-#include <qwt_plot_marker.h>
+#include <qwt_plot_textlabel.h>
 #include <qwt_plot.h>
 #include <qwt_plot_curve.h>
 #include <QFontDatabase>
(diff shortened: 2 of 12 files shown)


facontidavide:
Hi,

first of all, thanks for contributing.
I had a look to this changes and I have tried them with the fake data that I obtain running PlotJuggler with the option -t.

There are few things that I would like to change.

## Display (as text) of mean and standard deviation.

I would prefer to have this text in the upper left corner. And, of course, we need a good icon for the button, we should not use the same of the adjacent one.
I would also add max and min, but this is another story.

## Display (as curve transformation) of mean and standard deviation.

Here I am going to be a little less constructive (sorry in advance).
To me seeing a flat line with the mean or the standard deviation adds very little value to the end user, in particular when I can more easily read it as text. I am afraid that I must reject this particular change.

## FFT transform 

Conceptually I like this feature, even if personally I will never use it.
It I don't mistake, you are making the assumption that all the points in the curve are perfectly spaced in time, i.e. that the delta time.is assumption is in many cases false.
MAYBE you have data that fulfill this requirement, but I can not assume that all the user data will. 
This worry me because it creates in the user some expectations about the correctness of the displayed curve that may be false.
In other work, if PlotJuggler has the ability to display FFT, it must be the correct one !

What do you think?

CorentinChauvin:
For the display part, I totally agree. I did the curve transformation at first, but is not that necessary now.
For the FFT, the issue didn't even come into my mind before, to be honnest, and is surely problematic. Thank you for pointing it out!

I'll try to correct that within the next days.

facontidavide:
FYI. I didn't forget about this.
I will close the PR, but the good news is that I am thinking about a new class of Plugins called Toolboxes.

The first (and simplest Toolbox will probably implement thid functionality.

Cheers

CorentinChauvin:
Well, very nice. I'm very sorry for having given no news for this long
period.

Best,
Corentin

Le ven. 1 mars 2019 à 10:42, Davide Faconti <notifications@github.com> a
écrit :

PR #107, opened 2018-08-28
Fix dragging/deletion of hidden items

v-lopez:
If you have a list with multiple elements:
ABAR
FOO
ZBAR

And you filter by "BAR" you'll get a list of two elements, but if you shift-click on the first and the last, "FOO" will be selected as well, even though it's hidden. 

This patch should prevent FOO from being drag&dropped into the curve display and being deleted if it's hidden.


facontidavide:
Awesome. This is a recently introduced bug. Thanks for detecting and fixing it

PR #108, opened 2018-08-28
Custom calculations

1r0b1n0:
Hi,
Thanks for this great program !

This PR adds custom calculations / transforms on plots.
To open the dialog you can right click on a curve in the left listWidget, then select the "create math plot" option in the context menu.
Example ( where i extract the yaw euler angle from the 4 quaternion values ) :
quaternion_to_euler

Any javascript expression can be used for the calculations.

Another example (1st order IIR low pass filter) : 
add_low_pass1
add_low_pass2

_Some remarks :_

**Why javascript ?**
Using some compiled scripting language like [cling](https://github.com/root-project/cling) or lua would have been nicer, but Qt has native support for javascript (using QJSEngine). This makes the building and deploying a lot easier (I only added two words in the main CMakeList.txt).

**XY plot**
To make the right-click context menu on the FilterableListWidget work, I had to change the way to assign a plot to the X axis.
Instead of pressing right click and drag-and-drop, you just have to left-click drag the plot on the X axis.
I hope this change doesn't mind you !

**Streaming**
For the moment the calculations aren't refreshed during streaming, but in the context menu you can manually force a refresh of the expression.

Sorry for the big commit, this change was a little bit more complicated than expected !

### DIFF
--- a/include/PlotJuggler/plotdata.h
+++ b/include/PlotJuggler/plotdata.h
@@ -62,7 +62,7 @@ public:
 
   virtual ~PlotDataGeneric() {}
 
-  std::string name() const { return _name; }
+  const std::string &name() const { return _name; }
 
   virtual size_t size() const;
 
@@ -105,6 +105,7 @@ private:
 
 typedef PlotDataGeneric<double,double>  PlotData;
 typedef PlotDataGeneric<double, nonstd::any> PlotDataAny;
+typedef std::shared_ptr<PlotData> PlotDataPtr;
 
 
 typedef struct{
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ SET(CMAKE_AUTOUIC ON)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -g -fno-omit-frame-pointer")
 
-find_package(Qt5 REQUIRED COMPONENTS Core Widgets PrintSupport Concurrent Xml)
+find_package(Qt5 REQUIRED COMPONENTS Core Widgets PrintSupport Concurrent Xml Qml)
 find_package(Qt5 QUIET COMPONENTS Svg)
 find_package(Qt5 QUIET COMPONENTS WebSockets)
 
@@ -30,7 +30,9 @@ set( QT_LINK_LIBRARIES
     Qt5::Widgets
     Qt5::PrintSupport
     Qt5::Xml
-    Qt5::Concurrent)
+    Qt5::Concurrent
+    Qt5::Qml
+    )
 
 if( NOT Qt5Svg_FOUND)
     message(STATUS "Can't find Qt5Svg in your system. Have you tried [sudo apt-get install libqt5svg5-dev] ?")
--- a/package.xml
+++ b/package.xml
@@ -22,6 +22,7 @@
 <build_depend>ros_type_introspection</build_depend>
 <build_depend>qtbase5-dev</build_depend>
 <build_depend>libqt5-svg-dev</build_depend>
+<build_depend>qtdeclarative5-dev</build_depend>
 <build_depend>binutils</build_depend>
 
 <run_depend>rosbag</run_depend>
@@ -33,6 +34,7 @@
 <run_depend>ros_type_introspection</run_depend>
 <run_depend>qtbase5-dev</run_depend>
 <run_depend>libqt5-svg-dev</run_depend>
+<run_depend>qtdeclarative5-dev</run_depend>
 <run_depend>binutils</run_depend>
 
 <!-- The export tag contains other, unspecified, tags -->
--- a/plotter_gui/CMakeLists.txt
+++ b/plotter_gui/CMakeLists.txt
@@ -12,6 +12,7 @@ QT5_WRAP_UI ( UI_SRC
     filterablelistwidget.ui
     tabbedplotwidget.ui
     axis_limits_dialog.ui
+    add_math_plot.ui
     ../common/selectlistdialog.ui
     )
 
@@ -33,6 +34,8 @@ SET( PLOTTER_SRC
     filterablelistwidget.cpp
     tabbedplotwidget.cpp
     subwindow.cpp
+    add_math_plot.cpp
+    math_plot.cpp
     tree_completer.h
 
     mainwindow.h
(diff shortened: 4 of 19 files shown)


facontidavide:
Wow. This is very exciting. I am currently spending my holidays abroad and I am not able to properly review it. Once I will be back, I will be happy to look at it

1r0b1n0:
PR to add qtdeclarative (needed for QJSEngine) in rosdep : ros/rosdistro#18975

ahundt:
wow this looks amazing!

ahundt:
@1r0b1n0 I'm looking at the photos there and it is a little hard to tell what variables already exist, and what variables you will need to set for the output for things to work. Can the UI make that a little more clear or perhaps it is detailed in the help?

1r0b1n0:
Hi !
The code builds in Ubuntu 16.04 and 18.04 (and surely other distros).  The continuous integration fails because it is missing a dependency (apt install qtdeclarative5-dev), but I don't think I have access to change that.

I'm open to suggestions to make it more user-friendly !  I tried to explain in the "help" tab how everything works. Moreover in the curves tab you can double-click on a curve to add it's expression in the text editor.

1r0b1n0:
Great ! Enjoy your holidays !

include/PlotJuggler/plotdata.h
```diff
@@ -62,7 +62,7 @@ template <typename Time, typename Value> class PlotDataGeneric
 
   virtual ~PlotDataGeneric() {}
 
-  std::string name() const { return _name; }
+  const std::string &name() const { return _name; }
```
facontidavide: nice catch

facontidavide:
Hi,

first of all, let me tell you that it is great that you wrote this extension. I think this will be the most important feature for PlotJuggler in a long time.

For this reason, I want to be sure that we do it "right".

This is just a very preliminary feedback about the usability, not the implementation.

1) I don't mind the way you changed dragging the X axis of the XY curve. It may even be more intuitive for few users. But we must prevent the user from accepting more than 1 dragged curve (multiple curve selection).

2) My initial idea was to use Lua as scripting language (of course CLING was out of question). But since the Javascript parser is a natural dependency in Qt, maybe you do have a point.

3) We should work together to make this work with streaming, that is an important requirement for few people (including myself)

4) we must address the problem of time series which do not share the time axis, I think.

5) I find a little confusing the fact that __Linked data__ is the y of calc(x,y), where x is the time.
On the contrary, I like using __$$CURVE$$__ to get data from the right timeseries. 
On the other hand, we need the time points (x series) to be provided by a main curve... 
I don't know, I must think more about it.

This is just a preliminary feedback, I will think more about it during this week.

Cheers

facontidavide:
I am working on top of your version applying some changes. In particular, I am considering a different layout of the Left side widget

selection_153

I personally prefer to separate the user created MathPlots form the original data and also to give more control over the way they are refreshed.

Checking the box "Keep refreshing" the refresh is triggered everytime the original curves are updated.
What do you think?

1r0b1n0:
Hi,
I like your idea of putting the math plots in another layout !  The refresh could get a bit tricky if the user creates MathPlots based on other MathPlots.

I think adding streaming support shouldn't be very hard, I can try to code something when I find some free time. Ideally it be nice to implement it in an efficient way so that only new values are calculated. Morever it should be possible to keep the QJSEngine in memory when streaming is active, so that the "calc()" method doesn't have to be parsed and compiled at each update.

I agree the **linked data** is a bit confusing. Maybe we could rename **x** to **t** and remove the variable **y**, so the only way to access the "y" data is with **$$CURVE$$** ?

> we must address the problem of time series which do not share the time axis, I think.
I don't think I understand what you mean. Can you give an example ?

Cheers,

facontidavide:
> The refresh could get a bit tricky if the user creates MathPlots based on other MathPlots.

If I don't mistake, it is just sufficient to refresh them in the same order they were created.

> Morever it should be possible to keep the QJSEngine in memory when streaming is active, so that the "calc()" method doesn't have to be parsed and compiled at each update.

I guess this is the most sensible optimization and the lowest hanging fruit. Don't worry, I will do it.

> we must address the problem of time series which do not share the time axis, I think.
> I don't think I understand what you mean. Can you give an example ?

Maybe it is not a "problem", just something that must be clearly documented and must have a defined behavior.
Consider that we want to mix data from PLOT_A and PLOT_B and let's assume that the time t was taken from PLOT_A (it is the Linked data).

- PLOT_A might have 100 timepoints per seconds, whilst PLOT_B only 5 points per second (which do not correspond with any of the timepoints of the former.

- PLOT_A and PLOT_B can even be in two different (not overlapping) time ranges. What should we do in this case?

A typical example would be to use as PLOT_B data from __tf__.

facontidavide:
UPDATE

I am cherry picking the changes from your PR and merge them into branch __2.X__ . Therefore:

1) It will take me some more time
2) I will not merge the PR per-se.
3) There are still few things I want to improve.

1r0b1n0:
Ok great !

> Maybe it is not a "problem", just something that must be clearly documented and must have a defined behavior.
> Consider that we want to mix data from PLOT_A and PLOT_B and let's assume that the time t was taken from PLOT_A (it is the Linked data).
>     * PLOT_A might have 100 timepoints per seconds, whilst PLOT_B only 5 points per second (which do not correspond with any of the timepoints of the former.
>     * PLOT_A and PLOT_B can even be in two different (not overlapping) time ranges. What should we do in this case?
> A typical example would be to use as PLOT_B data from **tf**.

In the case where PLOT_A is at 5Hz and PLOT_B at 100Hz the calculations should work fine.
If the time ranges of PLOT_A and PLOT_B are not overlapping it won't work. For this situation we could maybe let the user generate an evenly spaced list of timestamps over the whole time range, like in this mock-up : 
image

facontidavide:
Ok, I am quite happy with the "merge". My updated code can be found in the branch https://github.com/facontidavide/PlotJuggler/tree/custom_math

I applied several changes and improvements to the code, being the most notable

- Split view of the left curve selector 

image

- Added the "Recently used functions" to the AttMathPlotDialog

image

- The user defined functions are saved and restored into the Layour files.

I will close you Pull Request. If you want to propose any additional change, please start from the head of the __custom_math__ branch.

facontidavide:
Well, it took a while, but finally this amazing feature has been merged to master.
Thanks again for contributing to PlotJuggler

1r0b1n0:
It works great, thanks !

facontidavide:
thanks you. may I have your real name to add it to the list of contributors? You may contact me by email if you prefer davide.faconti@gmail.com

PR #111, opened 2018-08-31
CMakeLists.txt add mac homebrew qt5 install directory

ahundt:
Fixes the build on MacOS, an extra qt path is added to the search path when cmake is run on an apple OS.


ahundt:
The appveyor issue doesn't look related to the code.

PlotJuggler opens OK, but it doesn't seem to load the plugins, when I go to `File > Load Data` I get this error:
```
No plugin was loaded to process a data file
```
Maybe the code isn't set up to load plugins ending in `.dylib`?

ahundt:
@facontidavide I just aded .dylib to the list of extensions to accept and now I can load CSV on macOS! Thanks for writing such good code that it only required less than 5 lines of changes to build on a new platform. 👍

PR #120, opened 2018-12-03
Add pal_statistics_msgs/Statistics rules

v-lopez:
Adding rules for visualizing properly the messages used by [pal_statistics](https://github.com/pal-robotics/pal_statistics)

I'm in the process of documenting it, you can see some info here: http://rosin-project.eu/ftp/pal-statistics

I'd like to recommend PlotJuggler as one the visualizers for the data, but it needs this rule to display properly.

We are logging over 2.5k variables from TALOS at 1000Hz in a  real time safe way, I'm sure you'll like pal_statistics.


facontidavide:
" I'm sure you'll like pal_statistics."

Estás hecho todo un comercial ;)

PR #136, opened 2019-02-10
Make default filename for tab image the tab name.

d-walsh:
This helps when saving multiple images of tabs.


facontidavide:
Nice

PR #137, opened 2019-02-11
Save all tab plots as images in a folder.

### LINKED ISSUE #128: Feature Request: Save muti-tab plots at once
Currently with multiple plots on a single tab the plots on that tab can be saved as a single image by right clicking the tab. 

It would be great if in a multiple tab layout you could export all the tabs at once, either to a multi-page PDF or multiple images (possibly using the tab names as the filenames).

d-walsh:
Extends https://github.com/facontidavide/PlotJuggler/pull/136 and closes #128 allowing a user to save all tabs as images in a selected folder.

### DIFF
--- a/plotter_gui/mainwindow.cpp
+++ b/plotter_gui/mainwindow.cpp
@@ -2358,3 +2358,30 @@ void MainWindow::on_actionSupportPlotJuggler_triggered()
 
     dialog->exec();
 }
+
+void MainWindow::on_actionSaveAllPlotTabs_triggered()
+{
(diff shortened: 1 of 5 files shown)


facontidavide:
thanks

facontidavide:
Unfortunately there are a couple of issues with this new feature.

1) If multiple tabs share the same name, the previous files with the same name are overwritten.
2) Zero warnings when files are overwritten.

I will fix these issues before the next release, but feel free to fix them yourself and send a PR ;)

PR #143, opened 2019-02-17
Allow to build the DataStreamClientSample on Linux

romainreignier:
Note that I am not sure if the `Windows` include is really needed in this file...
But I do not have a Windows machine to test.


facontidavide:
Thanks. This is a contributed plugin and I don't really use it

PR #152, opened 2019-03-01
Data streamer plugin that reads JSON from UDP

langrind:
Data streamer plugin that reads JSON from a UDP port and makes whatever comes in available to be plotted.

Also added tweaks to mainwindow.cpp to allow me to launch PlotJuggler from a shell script and begin presenting streamed data without any interaction from the user.


facontidavide:
Hi @langrind ,

first of all, let me say "thank you" for contributing to PlotJuggler. I like the idea of having this new plugin, buuuuuut...

The more people contribute to PJ, more I need to manage the fact that they don't add extra work to maintain the plugins. I already merged few PR that were "half baked" and I had to fix them (by the way, continuous integration fails.... you should check it out).

In the context of this particular plugin, I like it so much that I want to make it even better before merging it ;)

The idea is that given N serialization protocols and M transport layers, we may have potentially N*M plugins. of course we don't want that.

Serializetions __Protocols__ might be:
- JSON
- XML
- Google Protocol Buffer (requires schema)
- Flatbuffers (requires schema)
- etc...

__Transport__ Layers might be:

- UDP
- WebSockets
- MQTT
- ZeroMQ
- etc...

Therefore I would like to have, in particular at the level or Serializtion Parsing individual libraries.

I (we?) need to find a good way to allow the user to select individually Protocol and Transport.

facontidavide:
It would also be nice if you add a simple client application that publish the message, to make it easier for me to check if this work properly or not.

As a reference https://github.com/facontidavide/PlotJuggler/tree/master/plugins/DataStreamWebSocket/DataStreamClientSample

langrind:
Hi Davide,

Thanks so much for evaluating quickly and providing feedback. I really appreciate that. (Also really appreciate your excellent application! Easy to use, robust, readable code, all good!)

Please feel free to close my pull request without merging. I certainly didn't expect it to be accepted, since I just started using PlotJuggler last week, and still have a lot to learn about how it works internally, and how you like to do things.

In order to make transport and serialization plugins, I suppose you would need  an API. The current plugin layer interfaces to PlotJuggler via the DataStreamer class. I would think that would continue to be the API for a Serialization plugin. Then the Transport plugin would have a new API, that would allow a Serialization plugin to mate with a Transport plugin. 

I also suppose that creating this API and then supporting 4 or 5 transports would be enough to feel comfortable that the API is sufficient. 

I can imagine that there is simple Serializer config dialog that allows it to select any active transport. Serializers could save the binding in their XML config in the layout. You would have multiple transport instances (UDP port=10000 is one instance of UDP transport, UDP port=10001 is another instance of UDP transport) and so probably each transport instance needs a unique name.

I'd be happy to follow up on this with a prototype. Please let me know if you have ideas already, or any guidance. Or if you are already working on this, let me know what I can do to help - for instance I could try implementing additional transports if you already have an interface idea.

All the best,
Nik

facontidavide:
I took the first tiny step in the direction that interest both of us, creating this branch:

https://github.com/facontidavide/PlotJuggler/tree/parsers

The goal is, as I said in a previous message, to decouple **Protocols** from **Transport** layers.
For this reason I created a new type of plugin, the **MessageParser**, that basically represent a mechanism to implement a Protocol deserialization mechanism, independently from the actual transport layer used to deliver the message.

**MessageParser** are meta-plugins, because they are used by **DataLoader** and **DataStreamer** plugins.

facontidavide:
A question about your DataStreamJSON plugin...

Apparently you can parse very simple JSON files without any hierarchy, only key/value pairs. is that correct?

facontidavide:
I moved the discussion here #163, since it is unlikely that I merge this PR the way it is written.
Thank you for triggering this discussion.

By the way, do you work at PiaggioFastForward?

langrind:
That's right.

When I first made the plugin, I used a schema that I imagined would be
somewhat generic, with a label 'type' and a label 'instance' and a label
'value'. But I found that I had to have an entity that sat between my data
source (e.g. a drone) and PlotJuggler to massage the data. That entity was
useful for filtering and republishing the data, and also directing the data
to other live displays (e.g. bar widget, or a grid of latest-values).  Once
I realized that I was never going to want to live-graph data directly from
the source, I decided that the simplest scheme for the PJ plugin was best -
I just require that the middle entity make all the data that it sends to PJ
have nice sorting properties. The result is that the PJ JSON plugin can be
really simple.

I have been happy with this decision, but it suffers from the drawback that
it doesn't support a timestamp from the data source, so it's not perfect.

langrind:
No, I don't work at PFF, but I work with Alexey on a project that we use
PlotJuggler for

PR #157, opened 2019-03-14
ulog_parser: fixed parsing of array topics

RomanBapst:
**Problem**
I was not able to open log files which included arrays of nested topics. I think I found the cause why this was not working, however, the data of the nested topics is still not displayed correctly. There seems to be a miss-alignment between labels and data (only for that nested topic).

Unfortunately, I'm currently not able to share a log file that includes nested topics. (The one I was using was from a customer and I cannot share it here :-( ).
I will try to look for a public log. In my case the vehicle was logging the "esc_status" message which contains an array of "esc_report" message.

This is the error message that I saw when trying to parse the log file:
image

With this PR the log file is parsed and the message label appear, however, the data is not correct yet.
In the plot above I wanted to plot temperature but in fact the graph corresponds to the esc current.
image

@facontidavide Maybe you can help me get this right.

Signed-off-by: Roman <bapstroman@gmail.com>

### DIFF
--- a/plugins/DataLoadULog/ulog_parser.cpp
+++ b/plugins/DataLoadULog/ulog_parser.cpp
@@ -162,6 +162,12 @@ char* ULogParser::parseSimpleDataMessage(Timeseries& timeseries, const Format *f
 {
     for (const auto& field: format->fields)
     {
+        // skip _padding0 messages which are one byte in size
+        if (field.field_name == "_padding0") {
+                message += field.array_size;
+                continue;
+            }
+
         for (int array_pos = 0; array_pos < field.array_size; array_pos++)
         {
             double value = 0;
@@ -573,11 +579,7 @@ bool ULogParser::readFormat(std::ifstream &file, uint16_t msg_size)
             }
         }
 
-        if( field.type == UINT8 && field_name == StringView("_padding0") )
-        {
-            format.padding = field.array_size;
-        }
-        else if( field.type == UINT64 && field_name == StringView("timestamp") )
+        if( field.type == UINT64 && field_name == StringView("timestamp") )
         {
             // skip
         }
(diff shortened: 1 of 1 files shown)


facontidavide:
Thank you for reporting this issue. I do not have time to look at this right now, but I hope I will be able to fix it by the end of this month

RomanBapst:
@facontidavide I just found the problem and pushed a fix. The handling of the padding fields was not done properly. The esc_status messages has a padding fields before the nested message and that is what was causing the problem.
This works properly now.

plugins/DataLoadULog/ulog_parser.cpp
```diff
@@ -736,10 +748,15 @@ ULogParser::Timeseries ULogParser::createTimeseries(const ULogParser::Format* fo
     {
         for( const auto& field: format.fields)
         {
+            // skip padding messages
+            if (field.field_name == "_padding0") {
```
bkueng: You should skip all fields starting with `_padding`, see https://dev.px4.io/en/log/ulog_file_format.html#definitions-section

RomanBapst: @bkueng Thanks, will do!

RomanBapst: Done

facontidavide:
Thanks a lot. Awesome work.

plugins/DataLoadULog/ulog_parser.cpp
bkueng: I guess you forgot that one?

facontidavide: I took care of it

bkueng: Thank you.

PR #159, opened 2019-03-16
Make drag-drop work on MacOS

azaparov:
On MacOS there are several mime formats generated in addition to "curveslist", this fix will keep curves array with names collected instead of resetting it for each new mime format.


facontidavide:
Thanks!!!!!!!!!

PR #162, opened 2019-03-21
Use yyyy-MM-dd_HH-mm-ss name when saving a plot as image

### LINKED ISSUE #155: Save plots does not increment name
Hello, I find something not really convenient while saving plots as images.

Description:
The `Save all plot tabs as images` function always tries to save the first plot as `01_plot.png`. If that file already exists, it asks to overwrite it.

Possible solutions:
1. Ask the user to enter a filename, if it exists, ask to overwrite it
2. Increment the number is the file already exists

Many thanks!

bresch:
Fixes #155 

The name of the new image is `2019-03-21_10-41-22_01_plot.png` for example.


facontidavide:
thanks

PR #173, opened 2019-05-29
Read out of array

aeudes:
This Pr fix a possible invalid data access with back() on empty vector.
This could be triggered when used custom function in layout and stream in simulation (when layout is load before bag streaming start).


facontidavide:
Thank you very much!!!!

PR #175, opened 2019-05-29
Remove topic prefix

aeudes:
This pr add option to be able to remove prefix from bag and streaming topics. This is mostly useful for reused of layout when data are recorded with different namespace (/robot1,/robot2) ; but also to reduce topic length when compare different run of same algo (/algo -> /algo1 vs /run1/algo).
I am not totally sure that's the best way to do it (add prefix to layout when loaded should cover most of the usage of this patch).
The user can provide a list of prefix with ';' separator that are removed before adding the prefix name.

If you see interest on this PR, I can rebase it on the parsers branch.

### DIFF
--- a/compile_commands.json
+++ b/compile_commands.json
@@ -1,0 +1,1 @@
+../../build/plotjuggler/compile_commands.json
--- a/plotter_gui/mainwindow.cpp
+++ b/plotter_gui/mainwindow.cpp
@@ -1211,7 +1211,7 @@ bool MainWindow::loadDataFromFile(const FileLoadInfo& info)
 
             if( dataloader->readDataFromFile( &new_info, mapped_data ) )
             {
-                AddPrefixToPlotData( info.prefix.toStdString(), mapped_data.numeric );
+                AddPrefixToPlotData( new_info.prefix.toStdString(), mapped_data.numeric );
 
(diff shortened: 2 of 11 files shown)


facontidavide:
I should take my time to look at this...

aeudes:
I still feel that having control on topic prefix is still needed.
From my daily use of PlotJuggler, one usage of this pr is the fact that same algorithm are running on platform with different name/ROS prefix. To be able to use the same layout with all platform data, I need to change the prefix.
Same problem when I load multiple bag, having the control on prefix: allow me to assign meaning full name (better than A,B... in my opinion) or to not prefix topic at all if there are no topic collision or bag already contained prefixed topic.

Anyway as always thank-you for the great work on this very useful tool !

facontidavide:
since the tree view as been implemented, I think there is no more need for this.

Additionally, you can add multiple rosbag at once and the prefix will be updated automatically

ljburtz:
@facontidavide do you have a recommendation for being able to reuse the same --layout file for plotting the same topics but from two different robots -> eg the topics have a prefix of /robot1 and /robot2. 

Sorry for resurrecting an old thread, googling for a solution to this led me to this issue.

PR #179, opened 2019-06-02
Minor fixes

aeudes:
Three small fix:
1. the ref topic in custom function, could be change by the editor if this topic does not exist/is not loaded yet. The first commit add the ref in the combo box if not present.
2. the deleteDataMultipleCurves make a return if a curve if not found, I thinks it is more a continue here. As every curve should be deleted
3. the RosMessageParser::registerSchema did not return a value. This produce big mess with the optimiser ....

Also just as a side note:
the _builtin_parsers parsers objects are never deleted and I don't think they are automatic object.
https://github.com/facontidavide/PlotJuggler/blob/ace1da563ba09ce2363a08ec4e254a9c644bf498/plugins/ROS/RosMsgParsers/ros_parser.cpp#L19 
https://github.com/facontidavide/PlotJuggler/blob/ace1da563ba09ce2363a08ec4e254a9c644bf498/plugins/ROS/RosMsgParsers/ros_parser.cpp#L76

### DIFF
--- a/plotter_gui/mainwindow.cpp
+++ b/plotter_gui/mainwindow.cpp
@@ -866,7 +866,7 @@ void MainWindow::deleteDataMultipleCurves(const std::vector<std::string> &curve_
         auto plot_curve = _mapped_plot_data.numeric.find( curve_name );
         if( plot_curve == _mapped_plot_data.numeric.end())
         {
-            return;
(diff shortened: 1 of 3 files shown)


facontidavide:
Good catch!

PR #181, opened 2019-06-11
Virtual Override bug

SiyuChen1:
when I use `catkin_make` to build this project,an error occurs.


SiyuChen1:
Pure virtual functions must be implemented in the derived class. So I update and fix this bug.

facontidavide:
thanks

PR #183, opened 2019-06-26
Prompt user when one or more topics in the rosbag cannot be loaded

krixkrix:
* previous behaviour: the bag load was aborted
 * with this patch: prompt user whether or not to
   * skip this problematic topic and continue loading
   * skip this and other problematic topics that may arise when loading
   * abort the load

The patch addresses a real life example when some of the ros messages are created
from ROS-matlab, and therefore suffers from a bug related to a bad message description
E.g. see https://github.com/rosjava/rosjava_bootstrap/issues/16


facontidavide:
the issue with rosjava explains a few weird things that happened to people I know in the past...

Thanks for the addition.

PR #185, opened 2019-07-11
Dont take invisible curve into account for axis limit computation

aeudes:
A small change that did not take into account the invisible curve when compute the limit on x and y axis.
It also zoom out on x axis when visibility change is triggered.


aeudes:
I have rebased this one on master for my own use.
Feel free to close the PR if you will not merge it or give me some directions if correction are needed.

facontidavide:
thanks a lot for the PR

PR #200, opened 2019-09-11
data stream topic plugin

aeudes:
If multiple plotjuggler instance are running at the same time, only one could have a working data stream topic.


facontidavide:
looks good to me

PR #205, opened 2019-09-30
Add missed abseil_cpp depend

efernandez:
I had a compilation error that turned out to be related with an old version of `abseil_cpp`, but I found out that `PlotJuggler` didn't depend on `abseil_cpp`. This simply adds the missed dependency and also one missed header, although for it compiles find w/o this PR at the end.


facontidavide:
great, thanks

PR #211, opened 2019-10-03
Offset removal and aspect ratio interface

marceloparavisi:
Hello 

I changed the interface, so we can apply an offset to the series. 

I also added the function to disable the aspect ratio. 

Regards,
Marcelo

aspect
offset

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@
 
 /build*
 /*.user
+/.vscode
 
 site/images/
 /PlotJugglerInstaller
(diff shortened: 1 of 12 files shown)


facontidavide:
Hi,

Thanks for contributing.
I have few problems with this PR:

- how are limits supposed to work with streaming?
- I think the way the rectangle is calculated might be wrong.
- In general, what frustrates me, is that it make the UI and UX more complicated for anyone to solve the problem of a few.

facontidavide:
I decided to not add this functionality for he time being, even if I greatly appreciate your contribution

PR #220, opened 2019-10-27
Fix custom update

aeudes:
This PR fix two different problems related to custom function update.

=== When streaming (eda7c54) ===
If custom function is used, when streaming time is reset, the custom function is not updated anymore.
Step by step to make the problem happend :
- start streaming
- add custom function related to topic  and plot it
- start the bag 
- everything is ok
- restart the bag
- buffer is cleared and normal topic are display again but custom topic are not updated.

=== During data loading (e8649d7) ===
If data are loaded after custom function is defined, the custom function is not updated.
Step by Step:
- load layout with custom function define in it (keep placeholder).
- load data from bag.

### DIFF
--- a/plotter_gui/mainwindow.cpp
+++ b/plotter_gui/mainwindow.cpp
@@ -1904,13 +1904,14 @@ void MainWindow::updateDataAndReplot(bool replot_hidden_tabs)
         {
             _curvelist_widget->refreshColumns();
         }
-
-        for( auto& custom_it: _custom_plots)
-        {
-            auto* dst_plot = &_mapped_plot_data.numeric.at(custom_it.first);
-            custom_it.second->calculate(_mapped_plot_data, dst_plot);
-        }
     }
+
+    for( auto& custom_it: _custom_plots)
+    {
+        auto* dst_plot = &_mapped_plot_data.numeric.at(custom_it.first);
+        custom_it.second->calculate(_mapped_plot_data, dst_plot);
+    }
+    
 
     const bool is_streaming_active = isStreamingActive();
 
--- a/plotter_gui/transforms/custom_function.cpp
+++ b/plotter_gui/transforms/custom_function.cpp
@@ -44,7 +44,7 @@ CustomFunction::CustomFunction(const std::string &linkedPlot,
     _plot_name(plotName),
     _global_vars(globalVars),
     _function(function),
-    _last_updated_timestamp( - std::numeric_limits<double>::max() )
+    _last_updated_idx(0)
 {
 
     QString qLinkedPlot = QString::fromStdString(_linked_plot_name);
@@ -168,9 +168,9 @@ void CustomFunction::calculate(const PlotDataMapRef &plotData, PlotData* dst_dat
     }
 
     auto src_data_it = plotData.numeric.find(_linked_plot_name);
-    if(src_data_it == plotData.numeric.end())
+    if(src_data_it == plotData.numeric.end() || _last_updated_idx  == src_data_it->second.size())
     {
-        // failed! keep it empty
+        // failed =>  keep it empty ! or not nothing to do
         return;
     }
     const PlotData& src_data = src_data_it->second;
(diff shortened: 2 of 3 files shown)


facontidavide:
Hi,

I will review this as soon as possible and merge it ASAP.

facontidavide:
I am not sure I understand how the changes in  CustomFunction::calculate fix the issues.

Applying only the change in MainWindow::updateDataAndReplot seems to be sufficinet to fix the error

aeudes:
The bug is hiding in the fact that when time is reset, in this loop:
https://github.com/facontidavide/PlotJuggler/blob/1cd6efdc5c87c8fbdc6f0b8a6534471defcb9da5/plotter_gui/transforms/custom_function.cpp#L197-L206
the time index _last_updated_timestamp is not reset and cannot be less than source time. As the buffer is reset, dst_data is empty and _last_updated_timestamp is not updated.

I have changed from time constraint to index as it seems more clear to me to keep the transform progression this way. Maybe I miss something, I can reword the pull request without changing the way the indexing is done if you prefer. 
The other part of this change allows to avoid retrieving the auxiliary data, if there is no new source data (as the other commit force to call the update more often).

facontidavide:
Ok, so what about this code?

``` C++
    if (dst_data->size() == 0){
      _last_updated_timestamp = - std::numeric_limits<double>::max();
    }
    else{
      _last_updated_timestamp = dst_data->back().x;
    }

    for(size_t i=0; i < src_data.size(); ++i)
    {
        if( src_data.at(i).x > _last_updated_timestamp)
        {
            dst_data->pushBack( calculatePoint(calcFct, src_data, channel_data, chan_values, i ) );
        }
    }
```

Based on what you said, it might make more sense to move the update of _last_updated_timestamp before the for-loop

facontidavide:
actually I am realizing that based on the fact that timestamp must be monotonic, I don't even need to remember _last_updated_timestamp... i can write this code instead:

```C++
for(size_t i=0; i < src_data.size(); ++i)
{
    if( dst_data->size == 0 ||  src_data.at(i).x >=  dst_data->back().x)
    {
        dst_data->pushBack( calculatePoint(calcFct, src_data, channel_data, chan_values, i ) );
    }
}
```

aeudes:
Ok, this one is better. I am a bit bothered by the monotonic assumption on time as it is not true for time-series but the original code rely on it so....
I am still sure that the reset of src must be detected to clear the dst buffer (as it is not done when buffer is cleared and it is a push_back). I will make some test in this direction and proposed you an update of this PR.

``` C++
if (dst_data->size > src_data.size())
  dst_data->clear()

for(size_t i=0; i < src_data.size(); ++i)
{
    if( dst_data->size == 0 ||  src_data.at(i).x >=  dst_data->back().x)
    {
        dst_data->pushBack( calculatePoint(calcFct, src_data, channel_data, chan_values, i ) );
    }
}

````

facontidavide:
Confirm me if I am wrong, but the streaming bug happens ONLY when you use this option:

image

Am I right?

facontidavide:
Unfortunately this condition is triggered spuriously...

            if (dst_data->size > src_data.size())  dst_data->clear()

With the latest changes, both the original timeseries and the custom one need a "Clear Points" reset when you "rewind time".

I am going to push the changes myself and close this PR.

aeudes:
Sorry for long time delay in answering your questions and testing the new changes:
- The bug is trigger even if timestamp did not come from header.
- I have test your modification but when a partial buffer clear is done from ros streaming:
https://github.com/facontidavide/PlotJuggler/blob/e8daad7862e04e0e9dc862c864100aa52054ad50/plugins/ROS/DataStreamROS/datastream_ROS.cpp#L73-L90

I think your solution still not reset the custom data. As it is still good to have only partial reset and not silently clean other data source: I have made a proposal(0a6ad3a) in the updated branch to force clean of custom transform when reset time is detected.

- Also with the new loop when no monotonic stamp are processed, the custom will be processed only monotonically (contrary to previous behavior) and will possibly drop some data. This will cause custom time series to have different size/index than original series. I am not sure this is desired behavior. See 0ff62f7.

facontidavide:
I applied your changes. I think this should do. Thanks again!!!

PR #242, opened 2020-01-17
README.md update AppImage link to latest valid

dagar:
I initially thought this was a broken link, but on second glance it looks like the Github release is just missing for the newer tags after 2.4.1 (2.4.2, 2.4.3, 2.5.0). Feel free to discard...


facontidavide:
thank you, but this reminds me that I have to create the 2.5.0 instead ;)

PR #252, opened 2020-02-21
replace std::__cxx11::string with std::string

### LINKED ISSUE #251: Fails to compile: std::__cxx1::string does not name a type
In https://github.com/facontidavide/PlotJuggler/blob/869b250771153fe43f7062ea16a7c69f75356713/plotter_gui/transforms/lua_custom_function.cpp#L4 you use `std::__cxx1::string` instead of `std::string` which fails to compile on my machine.

In other parts of the application you use `std::string` therefore I assume it is an oversight and not intentional.

ftsell:
closes #251

### DIFF
--- a/plotter_gui/transforms/lua_custom_function.cpp
+++ b/plotter_gui/transforms/lua_custom_function.cpp
@@ -1,7 +1,7 @@
 #include "lua_custom_function.h"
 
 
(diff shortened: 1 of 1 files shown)


facontidavide:
thanks a lot for reporting this

Issue #262, opened 2020-03-26
Icons not getting installed


veimox:
In order for most of the desktops to recognize the icons they have to be installed in `/usr/share/applications`, `/usr/local/share/applications/` or `~/.local/share/applications/`. Similar for the icon.

When binarizing for ROS, the installation of the files [is not being done](https://github.com/facontidavide/PlotJuggler/blob/master/CMakeLists.txt#L142-L157). A way of fixing this could be just to move those install lines out of the if so they are always installed. We would also have to modify the install path to `/usr/share/` (absolute path).

The existing `.application` is calling `PlotJuggler` that as put in #261 might not give the desired behaviour (loading the pluggins) in the ROS version. This issues are not related but I think the ROS user would expect those plugins to work when using the icon too.

Issue #260, opened 2020-03-26
Scalable icon


veimox:
The existing icon is a 128x128. It will be simpler for desktop to utilize a `.svg` file instead

veimox:
Not sure if you have the original file... If not what [about something like this](https://www.svgrepo.com/svg/283627/line-chart-line-graphic)?

PR #263, opened 2020-03-26
installing icons in /usr/share and do it at any build type

veimox:
Tackles #262 but I would merge it along with a fix for #260


facontidavide:
already solved?

PR #265, opened 2020-03-26
Feature/scalable icon

veimox:
Tackles #260. This is just an example icon in svg. I can confirm that works correctly when installed! Should be merged in #263


facontidavide:
Your awesomeness level just increased +1

PR #266, opened 2020-03-28
installing plotjuggler bin in /usr/bin to make it globally accesible

veimox:
I forgot to push the commit that puts the binary in /usr/bin...

### DIFF
--- a/plotter_gui/CMakeLists.txt
+++ b/plotter_gui/CMakeLists.txt
@@ -88,7 +88,7 @@ if(COMPILING_WITH_CATKIN)
      )
     install(
         PROGRAMS scripts/plotjuggler
-        DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+        DESTINATION /usr/bin
     )
 else()
     install(TARGETS PlotJuggler DESTINATION bin  )


veimox:
up!

v4hn:
No!

That's what `${CMAKE_INSTALL_PREFIX}` is for if you don't use catkin to build.
I don't see why `mkdir build; cd build; cmake -DCMAKE_INSTALL_PREFIX=/usr ..; make; make install` should not do what you want.

You can't just assume the build has access to /usr/bin .
Also see #277 .

facontidavide:
yes, sorry @veimox , I should probably revert the changes, since they also break the creation of the AppImage.

A similar configuration works just fine with [rosbag_editor](https://github.com/facontidavide/rosbag_editor)

veimox:
@v4hn @facontidavide That's interesting... My idea is that the user has an icon that open the program. 

If you are against installing files outside of the catkin space I guess these lines have also to be removed:

https://github.com/facontidavide/PlotJuggler/blob/master/CMakeLists.txt#L154-L157

Maybe we can install everything in the catkin space and then modify the env variable that defines where icons are searched for. Do you have a proposal or is a lost cause?

v4hn:
Yes, as explained in the other issue I mentioned. :)

> Maybe we can install everything in the catkin space and then modify the env variable that defines where icons are searched for. Do you have a proposal or is a lost cause?

Well, you could install the files to `share/...` and if you build without catkin and the proper INSTALL_PREFIX, the files will be installed in the correct place.

Users could always expand their `$XDG_DATA_DIRS` to include any other path (like a ros workspace). You could also try to convince upstream `catkin` maintainers to extend this variable when you source a workspace. But that would not help you to get icons to show up outside a sourced workspace and invites wrong bug reports, so it's not likely to happen.

You could also detect user installations and install the files to [XDG_DATA_HOME](https://specifications.freedesktop.org/basedir-spec/0.6/ar01s03.html) (usually `~/.local/share/icons` and `~/.local/share/applications`).
Personally, I still consider that meddling with things you shouldn't really touch in an install to a specific PREFIX, but that's what it's for. Adding this as an optional feature might make sense.

https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html and
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
give the rest of the fine print.

veimox:
Sorry I didnt read the issue! for some reason i was getting redirected to #277  instead of #276 ...

In my opinion dropping this feature because it does not fit all the use cases is the wrong way of tackling the issue. I would rather create an if/else with either an existing global variable or with a new option. The installation in /bin or /share is as valid as the installation in /opt in my view for ROS package. At worst, this feature could be implemented for the non-ROS version part of the program -which turns to be the core of the program, ROS is an extension to the main program-.

If you dont see it that way, what do you think about having a flag or an option in the GUI to install the icons so the user can optionally install them?

v4hn:
> In my opinion dropping this feature because it does not fit all the use cases is the wrong way of tackling the issue

It's not about "fitting use cases". It's pretty much about conventions and expectations.

1. A standard interpretation is: `-DCMAKE_INSTALL_PREFIX` should install files only to this prefix
2. A system `/usr` installation can just install to the default `DATAROOTDIR` folder and will be picked up
3. The same works for any other path, except that users have to explicitly add the path to `XDG_DATA_DIRS`
4. `XDG_DATA_HOME` was introduced to provide a default user hook for user installations but violates 1.

If @facontidavide accepts 4. , that is a way to go if [`${CMAKE_INSTALL_PREFIX}/${DATAROOTDIR}`](https://cmake.org/cmake/help/v3.3/module/GNUInstallDirs.html) does not resolve to `/usr/share`.
The alternative approach is to *always* install the files to the path in the prefix and ask users to add it to their `XDG_DATA_DIRS` if they want it.

Either way, the current `desktop` file is broken, because it has to source the workspace if it was not installed to the system.

facontidavide:
I will investigate

PR #295, opened 2020-06-10
Pose with cov stamped

ClementAubry:
Hi Davide, Thanks for your incredible tool that is Plotjuggler!
I Add PoseWithCovarianceStamped message to pose_msg parser.


facontidavide:
ROS2 doesn't compile

facontidavide:
closing this because the way it is written it is breaking ros2 compilation

PR #316, opened 2020-08-11
Show warning message box if lines read from CSV are ignored.

maroc81:
Ran into a problem where a CSV file had ";" in a column so the entire file was ignored.  This will at least show a warning message box instead of silently ignoring invalid lines.


plugins/DataLoadCSV/dataload_csv.cpp
```diff
@@ -224,6 +228,14 @@ bool DataLoadCSV::readDataFromFile(FileLoadInfo* info, PlotDataMapRef& plot_data
     plot_data.numeric.clear();
   }
 
+  if (ignored_lines > 0 && lines_read > 0)
+  {
+    QString message = QString("Ignored %1 out of %2 lines read from file\n"
+                              "Valid delimiters are %3"
+                             ).arg(ignored_lines).arg(lines_read).arg(csv_separator.pattern());
```
maroc81, line 235: The delimiters will be shown in regex form but may want to show in a more user friendly form.

PR #317, opened 2020-08-12
Rosbridge integration

orshefi:
Hi Guys,

Spent the last couple of days writing this rosbridge WebSocket backend.

It's working, not airtight yet.

I'll keep improving it, I'll be happy to receive comments and help if some of you are able.

Tnx


facontidavide:
Hi,

I was planning a Websocket+JSON streaming plugin that is generic and not related to ROS at all.

How rosbridgecpp would compare to that? What is the problem / use case it will solve? Sorry for asking but this seems to be a very cool and important addition and I want to be sure that I understand how to use it, before merging it.

facontidavide:
furthermore, I would rather use QWebsocket and QJson to reduce the use of external libraries

orshefi:
Hi @facontidavide,

When I started this data streamer plugin I though I'll be using the ROS parsers to parse the messages from the WebSocket.
Eventually, this has proven to be unnecessary.

As you suggested, it's completely possible to decouple this module completely from both ROS and external libraries(use QWebsocket and QJson).

These are great ideas, I'll get started on fixing it.

facontidavide:
I can work on the json-to-plotjuggler part this weekend, maybe

orshefi:
Hi @facontidavide,

I had much progress with seperation of this feature from ROS and using Qt deps instead of third party.
There is no need for help at the moment, I'll post my updates in the next few days.

facontidavide:
Hi,

if I can make a suggestion about the way I imagine this.

I would use this library to be able to parse JSON, BSON, CBOR, MessagePack, and UBJSON with minimal effort.

https://github.com/nlohmann/json

Ideally, in the graphical part we should be able to specify the details about the Websocket server (port?) and the protocol.

The other important question is about the timestamp: is it the timestamp when the message was received or shall use the field "timestamp" in the message if present?.

I really want us to be aligned on these features ;)

orshefi:
Hi @facontidavide,

1. I don't really see a need at the moment to use JSON parser other than Qt built-in parser, do you?
2. I did create a small UI dialog for WebSocket address entry, I didn't allow method picking because the plugin was meant to rosbridge server interface, what other interfaces do you suggest?
3. I made it possible to enter the full address of the WebSocket, that's because I used ingress in Kubernetes to export my rosbridge WebSocket so I wouldn't like to limit the options.
4. Timestamp is a good question, at the moment I implemented the following - take timestamp from message header if exists, if not take message time of arrival.

facontidavide:
1. I think being able to parse more protocols with minimum effort is attractive, but we can postpone this.
2. I need to review what you did already
3. cool
4. It makes sense. We probably should add a checkbox "Use field timestamp is present"

facontidavide:
By the way, something that maybe it is obvious: the plugin we are discussing will replace the current (half-baked) one here:

https://github.com/facontidavide/PlotJuggler/tree/master/plugins/DataStreamWebSocket

What makes me unhappy about this plugin (contributed by an user) is the poor support for a proper protocol like JSON.

orshefi:
You are correct, another issue with that plugin is that plotjuggler is the ws server, that makes connecting with rosbridge server impossible.

The plugin I'm trying to add is a WebSocket client plugin, I want to be able to subscribe and publish in order to be able to get the full ROS development experience without direct access to ROS master.

orshefi:
Hi @facontidavide,

I updated the pull request with the new decoupled version.
It seems the tests are failing because QtWebsocket is not a part of the Qt bundle, can you add it?

 ```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake:28 (find_package):
    Could not find a package configuration file provided by "Qt5WebSockets"
    with any of the following names:
  
      Qt5WebSocketsConfig.cmake
      qt5websockets-config.cmake
  
    Add the installation prefix of "Qt5WebSockets" to CMAKE_PREFIX_PATH or set
    "Qt5WebSockets_DIR" to a directory containing one of the above files.  If
    "Qt5WebSockets" provides a separate development package or SDK, be sure it
    has been installed.
```

its a simple command: 
```
sudo apt install libqt5websockets5-dev
```

Other then that I currently have an issue I can't figure out, when not using the ROS header time stamp the samples are batchy.

orshefi:
Hi @facontidavide,

Added more functionality to enable publishing topics over websocket with ROS1.
Tested to be working with some topics, only sensor_msgs/Image is not working at the moment.

How do you feel about adding a video viewer to Plotjuggler? I think it'll upgrade the development and debug experience to see video & plots in sync.

facontidavide:
Video and image display is planned for plotjuggler 3,but it will take few months, since I am focusing on other features

orshefi:
@facontidavide,
I added basic video popups (modeless windows) on ws only, I had added a checkbox to allow popping up windows with the video, otherwise, it gets sent through the ws to the rosbridge server (Very slow at the moment).

I thought of a way to improve the way messages get sent through the WebSocket (so they won't overload the GUI thread) - split the WebSocket sending methods to a separate std::thread

facontidavide:
@orshefi I am excited about your contribution and this new feature, but PLEASE, keep this PR as small as possible and on point. The more things we add (like video) the less I am able to review it.

1263 files changed ????? I will not merge this, sorry.

I have some time to work on this, but I need your help:

- How can I test it? I am not familiar with rosbridge.
- Is it ready to be merged or still work in progress?

facontidavide:
There are few things that need to be improved to make this PR mergeable:

- remove all the websocketpp folder, since the [dependency can be installed with rosdistro.](https://github.com/ros/rosdistro/blob/20bfb944b18f51085480ae2df8b715ae40de50cd/rosdep/base.yaml#L4395)
- Remove the entire json folder and keep only the single header file.
- I noticed that you copy and pasted the files **dialog_select_ros_topics**. Is it possible to avoid that, moving the pugin into the ROS folder and linking to **commonROS** library?
- Please exclude from this PR everything related to imageview (that might be discussed in a future PR).

orshefi:
Hi @facontidavide, 

Sorry, it turned out to be such a long PR, most of the changes are just new libraries (websocketpp, nlohmann/json) I added in order to decouple dependencies and allow smother development experience.

The PR is stable and ready to merge, the reason I did it all in a single PR is that they are all related to the rosbridge capability.

### About Rosbridge
Essentially rosbridge is a websocket server that can be launched inside a ROS network, it allows websocket clients from outside ROS to subscribe/publish both services and topics : https://github.com/RobotWebTools/rosbridge_suite

The API is json api following this specifications: https://github.com/RobotWebTools/rosbridge_suite/blob/develop/ROSBRIDGE_PROTOCOL.md

### Changes
Most of the changes I've done were in two places:

1. I had added a new plugin called RosbridgeDataStream
   * It allows topic selection from the remote rosbridge server
   * It allows subscribing as the normal ROS plugin allows.
2. I had edited TopicPublisherROS
   * Added the capability to select publisher connection (ROS/WS)
   * Added the ability to pop up video for all sensor_msgs/Image topics when using websocket (sending these topics slows down the connection)

### Testing
1. Running rosbridge server is simple, you just install rosbridge suit ``` apt install ros-<distro>-rosbridge-suite ```
2. You run rosbridge using ``` roslaunch rosbridge_server rosbridge_websocket.launch ```
3. Then you can connect to it on ``` ws://localhost:9090 ``` from both publisher and subscriber


### Answers
While I was writing this comment you replied with some fixes you want done:
1. Removing websocketpp - don't you prefer it wasn't subjected to the user environment? if that's the guideline I'll do this for sure.
2. Didn't realize that JSON lib can run from a single header - will do.
3. Previously in this PR you requested to move this plugin outside of the ROS folder plugin because it does not depend on ROS libraries/compilation in any way, I think we should keep this that way, I can on the other end just link the files related to dialog_select_ros_topics from the ROS plugin directory and prevent the code duplication - do you agree?
4. Regarding the image view, I will split it into a separate PR. (Please let's discuss it soon as this is a very important feature for me).

facontidavide:
1. Ideally I prefered to use qt websocket, mostly thinking about Windows users. But since this plugin is by definition related to ROS, anything that is easily installable using rosdep, will be fine. Eventually this includes **websocketpp**- Therefore we can just add that dependency to **package.xml**

2. great

3. There was some misunderstanding from my side, since I didn't realized that anyway, this plugin is tightly related to ROS anyway. My bad.

4. We will discuss it in a different issue to prevent making this too noisy.

orshefi:
Great, I'll add websocketpp to package xml, remove anything but the single header from the json lib, move the streamer back to ROS plugin and split the video into new pr.

I'll get to do this probably tomorrow.

Have a good weekend.

facontidavide:
I guess I can use this to push info into the server, right?

https://roslibpy.readthedocs.io/en/latest/examples.html

I will also create a branch called "rosbridge". Please merge to that.

orshefi:
Yes you can use roslibpy to push info into server, easier way could be to use rostopic pub from cli.

Sure, I'll change the pr target.

orshefi:
@facontidavide,

I'm not familiar with how to add websocketpp to the package xml, can you give me a hint?

facontidavide:
https://github.com/facontidavide/PlotJuggler/commit/bdc19340e1d8f4e97b5e45e7d93d16ab514b7bed

this should be enough

orshefi:
Hi @facontidavide,

I did the following:
1. removed websocketpp from 3rdparty folder, linked it from the system.
2. deleted everything but the single json header as you said.
3. moved the rosbridge data stream inside the ROS folder.

I still had left to split video stream into separate PR, this I want to get inline with you on how to do that.
* Should I just copy the ui and header of the video popup without its integration into RosTopicPublisher?
* What branch should I split from?

Issue #335, opened 2020-10-01
NAN handling in ROS plugin 


farmange:
Hi there !
There is an issue when trying to import rosbag or stream a topic containing a float32 equal to NAN. 
This lead to a "Floating point truncated 1" error and nothing is imported.

facontidavide:
fixed

PR #336, opened 2020-10-01
Do not throw exception when trying to convert NAN

farmange:
Correct exception error when I want to import a rosbag containing float32=NAN (see issue #335). 
It also helps when live streaming ROS topics with same float32=NAN value.


facontidavide:
Thanks!

PR #339, opened 2020-10-15
pal_statistics fixes

v-lopez:
- Allow display of multiple pal_statistics_msg topics with same version
- Remove "values/" from pal_statistics variable names

The second change is cosmetic, but saves some screen space which is always nice.


facontidavide:
:+1:

asherikov:
Hi, it looks like this patch has been lost somewhere on the way. Is there a reason? Can I bring it back?

PR #340, opened 2020-10-15
Color of the XY Plot Symbol is set as color of the data

mucahitarduc:


mucahitarduc:
Ok then. I needed to distinguish between two separate XY plots to see the errors. Red color was dominant, so I changed that way.

facontidavide:
I appreciate that you wanted to contribute, though :smile:

PR #355, opened 2020-12-09
LabStreamlayer (LSL) plugin is developed.

csavur:
Dear Davide,

This pull request contains the LSL (https://github.com/sccn/liblsl) plugin. 
The plugin queries LSL stream in a local network, then behaves as a relay that passes data to PlotJuggler. 

We developed this on a weekend project. See below for the contributors:
- Celal Savur (@csavur)
- Shitij Kumar (@spk4422)
- Mazin Ali (@mazinMS)

Please let me if you have any questions.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/eclipse/paho.mqtt.c.git
 [submodule "3rdparty/liblsl"]
 	path = 3rdparty/liblsl
-	url = https://github.com/csavur/liblsl.git
+	url = https://github.com/PlotJuggler/liblsl.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/eclipse/paho.mqtt.c.git
 [submodule "3rdparty/liblsl"]
 	path = 3rdparty/liblsl
-	url = https://github.com/sccn/liblsl.git
+	url = https://github.com/csavur/liblsl.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "3rdparty/paho.mqtt.c"]
 	path = 3rdparty/paho.mqtt.c
 	url = https://github.com/eclipse/paho.mqtt.c.git
+[submodule "3rdparty/liblsl"]
+	path = 3rdparty/liblsl
+	url = https://github.com/sccn/liblsl.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.5)
 
 cmake_policy(SET CMP0048 OLD)
 PROJECT(plotjuggler)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,7 @@ include_directories(
     ./3rdparty/sol
     ./3rdparty/lua-5.4.1/src
     ./3rdparty/liblsl/src
+    ./3rdparty/liblsl/include
 )
 
 #########################  LIBRARY  ####################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -fPIC")
 ENDIF()
 
 find_package(ZMQ)
+find_package(LSL)
 
 include_directories(
     ${Qt5Core_INCLUDE_DIRS}
@@ -230,5 +231,11 @@ if( ZMQ_FOUND )
   add_subdirectory( plugins/DataStreamZMQ )
 endif()
 
+if( LSL_FOUND )
+  message(STATUS "LSL is found.")
+  add_subdirectory( plugins/DataStreamLSL )
+else()
+  message(STATUS "LSL is not found! please visit: https://github.com/sccn/liblsl")
+endif()
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.12)
 
 cmake_policy(SET CMP0048 OLD)
 PROJECT(plotjuggler)
@@ -103,7 +103,6 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -fPIC")
 ENDIF()
 
 find_package(ZMQ)
-find_package(LSL)
 
 include_directories(
     ${Qt5Core_INCLUDE_DIRS}
@@ -122,6 +121,7 @@ include_directories(
     ./3rdparty/qwt/src
     ./3rdparty/sol
     ./3rdparty/lua-5.4.1/src
+    ./3rdparty/liblsl/src
 )
 
 #########################  LIBRARY  ####################################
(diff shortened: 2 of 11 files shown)


mazinms (approved):

facontidavide:
First of all, let me tell you that this looks great, but I will need some time to properly review this. 

Also, you need to make yourself responsible for maintaining it :stuck_out_tongue: 

A problem I see is that the Debian package that is automatically built in the ROS farm will not have this dependency.

Therefore only people compiling from source can benefit from it.

A potential solution would be to add a git submodule, as I did with the MQTT plugin.

facontidavide:
Furthermore, how can I test it?

csavur:
> First of all, let me tell you that this looks great, but I will need some time to properly review this.
> 
> Also, you need to make yourself responsible for maintaining it 
> 
> A problem I see is that the Debian package that is automatically built in the ROS farm will not have this dependency.
> 
> Therefore only people compiling from source can benefit from it.
> 
> A potential solution would be to add a git submodule, as I did with the MQTT plugin.

I will be responsible for maintaining the source code since I am using it. I will add "LSL"  as a submodule no problem at all.

facontidavide:
If you have a loot to other streaming plugins (MQTT, websockets, UDP), you will notice that I include a python file with a dummy data publisher. We should probably do the same, since apparently, we can easily get **pylsl**.

csavur:
> Furthermore, how can I test it?

You can install this package "https://github.com/labstreaminglayer/liblsl-Python"

Then you can create a dummy stream by "**python -m pylsl.examples.SendData**". 

then it should just show up in the local network.

Let me know if you have any question.

csavur:
> If you have a loot to other streaming plugins (MQTT, websockets, UDP), you will notice that I include a python file with a dummy data publisher. We should probably do the same, since apparently, we can easily get **pylsl**.

Will do.

facontidavide:
One more question. Is there a list of supported hardware devices, so that we can brag about it with our users :wink: ?

csavur:
> One more question. Is there a list of supported hardware devices, so that we can brag about it with our users  ?

LSL is a really popular library that is used in biological experiments. 

Many devices are supporting the protocol. 

Here is the list of some of the devices,

https://github.com/sccn/labstreaminglayer/tree/master/Apps

Celal

csavur:
Just to make sure,

I am going to add liblsl as a submodule similar to mqtt, is that right?

facontidavide:
yes. We need to be sure that no library or include file is installed. I am having this problem with the MQTT library, actually, and it is super annoying.

csavur:
> yes. We need to be sure that no library or include file is installed. I am having this problem with the MQTT library, actually, and it is super annoying.

I had to increase the CMake version to 3.12, is this fine?

facontidavide:
Unfortunately it is not. I need to compile with Ubuntu 18.04 and the default version is 3.10.2

csavur:
> Unfortunately it is not. I need to compile with Ubuntu 18.04 and the default version is 3.10.2

liblsl requires 3.12, what should we do?

csavur:
> Unfortunately it is not. I need to compile with Ubuntu 18.04 and the default version is 3.10.2

I tried to reduce liblsl CMake version to 3.10, it compiles just fine. What we can do is i can fork liblsl to my account and add that project as a submodule so we can reduce the Cmake version issue. 

What do you think?

facontidavide:
it makes more sense, then to fork it on the plotjuggler account... https://github.com/PlotJuggler/liblsl

The problem is that it does NOT compile on Ubuntu 18.04 I am trying but unsuccesfully.

csavur:
> it makes more sense, then to fork it on the plotjuggler account... https://github.com/PlotJuggler/liblsl
> 
> The problem is that it does NOT compile on Ubuntu 18.04 I am trying but unsuccesfully.

I made two changes and send you pull request. If you accept the two pull request you will be able to compile on the ubuntu 18.04. I am using the same it works now. info_receiver.cpp was missing a include.

facontidavide:
There must be something different between your computer and mine, because mine is not compiling.

csavur:
> There must be something different between your computer and mine, because mine is not compiling.

I added the forked version as a submodule and add a dummy python file. I do not know why AppVeyor is failing. It looks like something else. 

Let me know if you have any questions.

facontidavide:
can you confirm me which version of Ubuntu, GCC and cmake you are using?

csavur:
> can you confirm me which version of Ubuntu, GCC and cmake you are using?

Well, I just figure out. When I compile from the terminal it gives me an error as well. When it uses  CMake 3.10.2 it fails. 
However, when I am using Qt Creator which is using  CMake 3.19.0, and it compiles just fine. 

g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

Sorry, I did not realize Qt Creator using a different one.

csavur:
Hello Davide,

What should we do? When are you planning to upgrade the CMake version? Or any other alternative solution.

facontidavide:
>  When are you planning to upgrade the CMake version?

In 2023, EOL of Ubuntu 18.04

Clearly, we need to find a different solution :wink:

csavur:
I will spend some time to see if I will make the project build with the CMake 3.10.2. I will let you know as soon as I have something.

facontidavide:
Have a look at this.

I did it in a quick and dirty way, I don't care about windows for the time being.

I also removed everything related to shared library and installation, since that may create problems down the road.

csavur:
> Have a look at this.
> 
> [PlotJuggler/liblsl@8bc4f56](https://github.com/PlotJuggler/liblsl/commit/8bc4f560ffdc2873202d87e17922ecaa9ec6a8f5)
> 
> I did it in a quick and dirty way, I don't care about windows for the time being.
> 
> I also removed everything related to shared library and installation, since that may create problems down the road.

It did compile but I guess there is a problem with the CMake File. It crashes when I try to connect a stream here is the error.

./plotjuggler: symbol lookup error: /home/celal/workspace/PlotJuggler/build/bin/libDataStreamLSL.so: undefined symbol: lsl_library_info

csavur:
Ok, I fixed it. everything is working now. I will send a pull request on liblsl then it is all set. Can you please accept the PR on liblsl/

csavur:
Hi Davide,

I made the final changes, and it should compile with your system. Can you review the PR and let me know any changes you want me to add.

csavur:
> Have a look at this.
> 
> [PlotJuggler/liblsl@8bc4f56](https://github.com/PlotJuggler/liblsl/commit/8bc4f560ffdc2873202d87e17922ecaa9ec6a8f5)
> 
> I did it in a quick and dirty way, I don't care about windows for the time being.
> 
> I also removed everything related to shared library and installation, since that may create problems down the road.

Did you have a chance to review the code and able to run the plugin?

facontidavide:
not yet. I will eventually. I am very busy with other things right now

facontidavide:
I have tested and it seems to work.

In terms of user-interface, it is a bit sloppy. For instance, it will not refresh the list of available channels. I am fixing that myself, but please, in the future, give some important to the UI, not just the functionality.

csavur:
> I have tested and it seems to work.
> 
> In terms of user-interface, it is a bit sloppy. For instance, it will not refresh the list of available channels. I am fixing that myself, but please, in the future, give some important to the UI, not just the functionality.

Yes, I did not refresh the stream. I will be more careful. I am glad it is working. Please assign me any bug that may happen in the future.

facontidavide:
FYI: https://github.com/facontidavide/PlotJuggler/commit/111f40eec38b8df319ae95fac912d97a7a734689

> Please assign me any bug that may happen in the future.

That is easy: can you please check that this works in Windows too? I think we need to fix the CMakelists.txt of **liblsl**

csavur:
> FYI: [111f40e](https://github.com/facontidavide/PlotJuggler/commit/111f40eec38b8df319ae95fac912d97a7a734689)
> 
> > Please assign me any bug that may happen in the future.
> 
> That is easy: can you please check that this works in Windows too? I think we need to fix the CMakelists.txt of **liblsl**

It is really nice that you added a timer to check every second. I have not tested on Windows yet, I am mainly working on Ubuntu and Mac. I will test on windows when I have time. As you said probably we will need to update CMakefile.

By the way, once we upgraded CMake 3.12 or higher we can depend on liblsl repository to get updates and bug fixes.

PR #358, opened 2020-12-10
Template parameter named ItemType collides with enum called ItemType

gabm:
MSVC fails to compile `SortedTableItem<QTreeWidgetItem>`.

If we use that definition, the class QTreeWidgetItem becomes a baseclass of SortedTableItem. That also inherits an enum called "ItemType" (which is the same as the template name). MSVC stumbles and spits out compilation errors :facepalm: . A simple solution is to rename the template parameter :facepalm: :facepalm:

### DIFF
--- a/src/curvelist_view.h
+++ b/src/curvelist_view.h
@@ -18,11 +18,11 @@
 
 class CurveListPanel;
 
-template <typename ItemType>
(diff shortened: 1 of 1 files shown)


facontidavide:
awesome

PR #376, opened 2020-12-28
Fixed message_parser plugin loading segfault

### LINKED ISSUE #375: MessageParserCreator plugin loading fails on name()
When mainwindow.cpp attempts to load a MessageParserCreator plugin, it attempts to call name() on the wrong object.

Lines 463-465:

> else if (message_parser){
>   plugin_name = streamer->name();
> }

jlmcmchl:
Fixes #375


facontidavide:
Thank you a lot for spotting this!

image

PR #386, opened 2021-01-15
Allow custom function return multiple points

### LINKED ISSUE #385: Allow custom lua functions to return an arbitrary number of samples
For now, Lua functions must return exactly one point. It would be great to be able to return any amount of point (including 0).

My use-case is this-one: I want to reproduce the [step](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.step.html) behavior of pyplot, and for this I could create a custom function which would return a second sample just before the new one, that is equal to the previous value.

Hugal31:
Remplace `CustomFunction::calculatePoint` by `CustomFunction::calculatePoints`, which takes a caller-allocated vector reference in parameter (instead of returning one, to reduce the number of allocations), and can therefore insert multiple points.

Lua functions can now take advantage of this by returning an array of two-number arrays:

```lua
-- Still valid
return value
return time, value
-- New syntax
return {{time - 0.01, value * 0.1}, {time, value * 1.1}}
```

BREAKING CHANGE: `CustomFunction::calculatePoint` is replaced by `CustomFunction::calculatePoints`

Closes #385.


facontidavide:
Nice, I will test it over the weekend.
We should probaly add some documentation / tutorial

Hugal31:
I added a small text in the cheatsheet. However you must scroll to see it, I'm not sure how to improve that.

cheatsheet

facontidavide:
mmh, no, I prefer a pop up dialog, please do not change the QLabel

Hugal31:
Removed.

PR #392, opened 2021-01-22
Apple Mac M1 build fix.

devnexen:
backward-cpp dependency fix for ARM 64 backport, wrong access to PC register.


facontidavide:
thanks for contributing

PR #393, opened 2021-01-25
Update udp_server.cpp

alkaes:
Fixes random corruptions of UDP Json messages (garbage collector related?)


facontidavide:
thanks

PR #396, opened 2021-01-28
Updated support for windows build + installer

alkaes:
Added win32build.bat batch file for easy windows builds (need to update QT path variables inside to correct ones in case it does not work)
This patch works great for me. I have a prebuilt installer on my own fork of this great project: [PLotJuggler v3.07 (Windows)](https://github.com/alkaes/PlotJuggler/releases/tag/v3.0.7)


facontidavide:
looks cute. Might delete later

PR #406, opened 2021-02-12
Use format string when time index is not a number

### LINKED ISSUE #403: [Feature Request] User defined parsing of timestamp field [use QDateTime::fromString()?]
Would be nice to have the option for user defined parsing instructions of the timestamp field. QDateTime::fromString() is relatively feature full., and the user can be prompted to provide a string compatible with that method. Currently QString::toDouble() is used which is very limiting. This should be a relatively modest change which would allow for wider adoption.

faisal-shah:
This implements a user defined format string to parse the time index field of a csv file. This becomes active when the field is not a number. This could obviously have deeper integration, such as storing the format string in a layout, or something of the sort. Consider this a usable proof of concept.

fixes #403


facontidavide:
looks good. Thanks!!!

PR #407, opened 2021-02-16
XY curve markers: fixed colors and removed ghosts symbols

blanzer:
**Issue**: The xy-curves had always red symbols as markers, which gets a bit confusing if there are multiple curves in one widget.

**Changes**: 

1. Marker color is now set the line color (either automatically assigned if line is added or to the newly selected color in plotwidget_editor)
2. Additionally, removed the tracker redrawing in xmlLoad since this resulted in ghost-markers at the origin which should be there. I was not so sure about this, but I tested it and it appears to be the fix for it


facontidavide:
you are the third pñerson that tries to change by beatiful (intentionally) red dot.

I give up. :smile:

PR #420, opened 2021-03-27
Add dataload can

ozzdemir:
With dataload_can plugin, log files created by candump -L can be visualized using PJ.
Currently, importing .log files trigger another file dialog for choosing the CAN database (.dbc).


facontidavide:
Hi, first of all, I used CAN a lot in the past and I am excited to get this into PlotJugler. But I decided that I want future plugins to be in separate repos here:

https://github.com/PlotJuggler

I also created a placeholder for you: https://github.com/PlotJuggler/plotjuggler-CAN-dbs

I will be happy to make you the ONLY responsible and administrator of this repo :) 

I suggest you to add a README with some screenshots and howto for anyone using the plugin.

What do you think?

facontidavide:
note that all the CI pipeline fails

ozzdemir:
Putting plugins in another repo would be quite nice due to some dependencies being not met by the default configuration. Actually, I checked where to put some instructions(), etc. but couldn't find a proper location previously, so I would be happy. Additionally, I was able to work the CAN streamer using the same dbcppp library with Qt5SerialBus, but it does not exists in regular installation (at least, in ubuntu 18.04). What would be the best approach? (Also containing the question of should I add it also to the same repository :smile: )

facontidavide:
I updated (finally) this repo: https://github.com/PlotJuggler/plotjuggler-sample-plugins

Here you will see an example of a "stand alone" plugin. Focus in particular on the CMakeLists file.

The repo is yours, do whatever you want, but please make a good README to make the life easier to your users :)

ozzdemir:
Awesome, will adapt as soon as possible :)

ozzdemir:
Development will be continued in the https://github.com/PlotJuggler/plotjuggler-CAN-dbs, so closing.

PR #424, opened 2021-04-19
Prefix of Logs

canpinaremre:
Prefix of logs are not longer A,B,C,D...
Logs' names are now their prefix.
This way it is easy to follow graps.


facontidavide:
I prefer to use  QFileInfo( info.filename ).baseName()

Cherry picking your changes.

PR #459, opened 2021-06-14
Add quaternion to Euler conversion snippets

bresch:
@facontidavide This have been really useful to me in the past to inspect PX4 (`.ulg`) logs and will be really useful to other users. Instead of having to share a `.xml` gist to everyone, I'm wondering if you would be OK to add these functions to the default snippet set.

Summary:
Add 3 functions to convert a Hamiltonian attitude quaternion to its Euler (Tait-Bryan 321/yaw-pitch-roll) representation

Result:
DeepinScreenshot_select-area_20210614172642

### DIFF
--- a/plotjuggler_app/resources/default.snippets.xml
+++ b/plotjuggler_app/resources/default.snippets.xml
@@ -55,4 +55,51 @@ end
 return value - first_value</function>
     <linkedSource></linkedSource>
   </snippet>
+  
+  <snippet name="quat_to_roll">
+    <global></global>
+    <function>w = value
+x = v1
+y = v2
+z = v3
+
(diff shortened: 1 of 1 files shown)


facontidavide:
I think it is nice, but I think there is a better way to do this.

I am plannignto build a new set of Plugins called "Toolbox". Quaternion to Euler builder would be actually the simplest I can think of and a good example.

facontidavide:
No reason not to merge this, though! Thanks a lot

bresch:
Thanks for merging it. I'm looking forward to seeing the "Toolbox" plugin!

mengchaoheng:
@facontidavide is publish now?  I need   quaternion to its Euler.

PR #483, opened 2021-07-21
Proper cmake way to find xcb

Tobias-Fischer:
We found that Plotjuggler does not compile in the RoboStack project for ros-galactic, as xcb include directories are not found. This patch fixes the compilation, using new style cmake targets.


facontidavide:
This change breaks compilation in the **CI**

facontidavide:
Closing until CI complains

PR #485, opened 2021-07-28
[WIP] Merge data instead of replacing it

TonyWelte:
This changes the way new data is added. New data with an already known name is merged with the old data (if the user wants to keep it).

This enables multiple bags recorded over the same timespan to be replayed (as "plotjuggler::rosbag1::consecutive_messages" can be merged).

It also enables to load simultaneously data recorded over a different timespan.

Relates to PlotJuggler/plotjuggler-ros-plugins#27

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -1150,15 +1150,16 @@ void MainWindow::importPlotDataMap(PlotDataMapRef& new_data, bool remove_old)
     AddToDeleteList( _mapped_plot_data.numeric, new_data.numeric );
     AddToDeleteList( _mapped_plot_data.strings, new_data.strings );
 
-    if (!old_plots_to_delete.empty())
+    if (!_mapped_plot_data.numeric.empty() || !_mapped_plot_data.strings.empty()|| !_mapped_plot_data.user_defined.empty())
     {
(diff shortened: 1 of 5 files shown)


facontidavide:
Thanks for taking the time to contribute to plotjuggler, but I am afraid that it may introduce some unexpected behavior.

I would feel more comfortable if a third option "Merge data" is available in 

          QMessageBox::question(this, tr("Warning"), tr("Do you want to remove the previously loaded data?\n"),

TonyWelte:
I agree adding a third option is best.
Now the "Yes" button is the same as before, the "No (remove duplicates)" should be the same as the old "No" button, only the "No (merge duplicates)" adds a new behavior.

bonkuraps:
@facontidavide would there still be a need for this feature? What do you think?

facontidavide:
There are a number of potential drawbacks that I need to analyze. In general, there is a 80% probability that this is not a good fit for PJ, at this moment of time.

This doesn't mean I don't appreciate your contribution, but i must be mindful about the feature I merge.

bonkuraps:
Ok, Is there anything that I might help with? Do you see any tickets that might be worth going through.

facontidavide:
Thanks a lot for the help offer!!!
Maybe #724 , #705, #681

PR #503, opened 2021-09-01
Fixed small issues with PreferencesDialog and use of skin to set the main window title 

GerardoPardo:
Fixed wrong slot name on PreferencesDialog and moved skin-based setting of MainWindowTitle after the setupUi() call


facontidavide:
thanks!

PR #512, opened 2021-09-07
Macos

GerardoPardo:
Modifications to compile and display properly on macos, tested on macOS Catalina version 10.15.7 with Qt version 5.15.2
Modificatios to improve darkmode. Mostly icons.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,1 @@
 
-[submodule "plotjuggler_plugins/DataStreamRTI"]
-	path = plotjuggler_plugins/DataStreamRTI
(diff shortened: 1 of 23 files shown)


facontidavide:
unfortunately the changes in CMakeLists files break compilation in ROS2. 
I need to investigate further...

GerardoPardo:
I tested your latest commit https://github.com/rticommunity/PlotJuggler/commit/e666333faf79d30021bf25865f6e983f9deb7e75 on the mac  and it compiles and runs well. 

It seems that the target_link_libraries addition is only needed if the add_library specified the SHARED option...
So this CMakeLists works for Mac as well.

PR #546, opened 2021-10-21
Add factors for speed transformation and invert by minus 1

pkess:
Hello,

i think the dialog with transformation is very usefull, but factors to convert m/s to km/h and to invert by -1 or change back to 1 would be very usefull, to. So i added them here. Would be happy to get this merged and available for the windows version.

Unfortunatelly i was not able to compile and test this cause there is no compiler set up on my pc.

### DIFF
--- a/plotjuggler_app/transforms/scale_transform.cpp
+++ b/plotjuggler_app/transforms/scale_transform.cpp
@@ -19,6 +19,30 @@ ScaleTransform::ScaleTransform() : _widget(new QWidget()), ui(new Ui::ScaleTrans
     emit parametersChanged();
   });
 
+  connect(ui->buttonFactorKmhMs, &QPushButton::clicked, this, [=]() {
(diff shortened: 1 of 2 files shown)


facontidavide:
Hi,

First of all, thank to take the initiative to create a PR to make plotjuggler better.

Unfortunately, my philosophy is that it is hard to take into account everybody 's need, and I prefer to keep a clean interface that take into account 70 % of the use case, rather than a more complex one that makes easy 80% of them.

Thanks anyway!

pkess:
Hello, thank you for your fast response.
While those constants may not be relevant for most of the users it may be helpful for most users if they can define their most important constants in a config file or the settings.

So would you agree in some settings entry do define additional buttons? If you do i can make a new Pr for this.

facontidavide:
it is not a bad idea. Let me think about it about how that can be done.

What a bout a "last used" button?

PR #573, opened 2021-12-14
Added libprotoc-dev to the apt install package list

martonmiklos:
To compile the Protobuf parser plugin I needed to install  libprotoc-dev on Ubuntu 20.04, so it might be useful to note this in the Readme.md


facontidavide:
thanks

PR #597, opened 2022-02-04
Fix MutableTimeseries shadowed by MutableScatterXY

simchanu29:
I noticed on the last commit of the main branch I weren't able to use MutableTimeseries. It seemed after some investigations it was always overshadowed by MutableScatterXY. 

Way to reproduce the bug : 
1. Use Reactive Function Editor
2. Create a new MutableTimeseries
3. The created timeserie is a scatter plot.

I guessed it was because new_usertype in sol overwrite last type if a previous type with a different name but same base class has already been declared. So I used inheritance to declare two subtypes `CreatedSeriesTime` and `CreatedSeriesXY` deriving from `CreatedSeriesBase`. 

Side note I also added lua math that I think was missing.

### DIFF
--- a/plotjuggler_base/include/PlotJuggler/reactive_function.h
+++ b/plotjuggler_base/include/PlotJuggler/reactive_function.h
@@ -5,7 +5,9 @@
 #include <sol/sol.hpp>
 
 class TimeseriesRef;
-class CreatedSeries;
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks a lot. I should release this fix soon

PR #601, opened 2022-02-07
Timestampfield

ngpbach:
I modified the checkbox in streaming plugins to use timestamp field, so that it can use a different field. This is useful to me and I only tested with UDP and ZeroMQ. Using a field with bogus data as timestamp can cause buggy behavior, but that's probably some core function I'm not sure I have the ability to fix.

PlotJuggler


facontidavide:
I think it is a good change, but I should try to make this a little more resilient.

I will have a look at it this week, thanks

Issue #619, opened 2022-03-21
No ZMQ plugin


MrShedman:
The ZMQ plugin has gone missing from latest version.

On either ROS1 or ROS2 I have no ZMQ plugin. I can also see in the latest appimage there is also no libDataStreamZMQ.so. Why is this missing?

My system:
- Ubuntu 20.04
- ROS1/2 (Noetic/Foxy)

facontidavide:
if you don't have ZMQ installed in your system, the related plugin will not be built

facontidavide:
I think a recent change should fix this

PR #626, opened 2022-04-15
Fix finding ZMQ library for DataStreamZMQ on Ubuntu

Inquisitive-ME:
I have only checked on Ubuntu, but ZMQ libraries are not found even when they are installed. I believe this would fix this issue  #619


Inquisitive-ME:
Closing in favor of https://github.com/facontidavide/PlotJuggler/pull/627

PR #627, opened 2022-04-19
Build ZMQ library instead of finding it on system

Inquisitive-ME:
Finding zmq library was broken on Ubuntu and also seems broken on Windows. It seems like it would be easier to just build it.

### DIFF
--- a/plotjuggler_plugins/DataStreamZMQ/CMakeLists.txt
+++ b/plotjuggler_plugins/DataStreamZMQ/CMakeLists.txt
@@ -1,26 +1,29 @@
-
-find_package(ZeroMQ CONFIG QUIET)
-
-if(ZeroMQ_FOUND)
-    include_directories(../ ${ZeroMQ_INCLUDE_DIR})
-
-    add_definitions(${QT_DEFINITIONS})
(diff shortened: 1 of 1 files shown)


facontidavide:
Couple of questions before merging:
- will this fail to build if there is no internet connection?
- can we use caching as descrived [here](https://gitlab.kitware.com/cmake/cmake/-/issues/18257)?

Inquisitive-ME:
The build will hang if there is no internet connection. I could not find a clean way for fetch_content to handle this. I think if this is a concern the other option would be to just add zmq as a submodule to the project.

Caching is already part of `FetchContent_Declare` so once CMake has been run it will put the contents in the build folder and will not need to download it for each build. There are additional options in the [documentation](https://cmake.org/cmake/help/latest/module/FetchContent.html). Notably to not fetch on updates `FETCHCONTENT_UPDATES_DISCONNECTED` or set a source directory `FETCHCONTENT_SOURCE_DIR_`

facontidavide:
I decided no to merge this, even if I am grateful that you spent time to contribute to PlotJuggler.

PR #649, opened 2022-05-11
add STATUS to CmakeLists.txt message() to avoid 'message called with incorrect number of arguments'

lucasw:
cmake 3.22.1 errors on this


facontidavide:
thanks

PR #657, opened 2022-05-28
Make CSV time column choice persistent.

agoessling:


facontidavide:
I prefer another implementation. https://github.com/facontidavide/PlotJuggler/commit/f8ee1fa18f2e769c216fae320a3994713cdb44a0

Thanks anyway for the idea :smiley:

PR #665, opened 2022-06-11
New widget for Lua script editing

Bartimaeus-:
Replace simple QPlainTextEdit in Lua script editors with widget from [Qt Code Editor example](https://doc.qt.io/qt-6/qtwidgets-widgets-codeeditor-example.html) to add line numbering and highlighting. The new widget inherits from QPlainTextEdit and original functionality is untouched.

Places updated with new widget:

- Custom Lua script editor (single and batch)
- Reactive Script Editor

Demo video below:

https://user-images.githubusercontent.com/2954254/173200613-2bff3d7b-1410-4b9d-bb2f-c05e53bc3273.mp4

A Qt license info dialog was also added per the Qt BSD license for the widget:

https://user-images.githubusercontent.com/2954254/173200725-36a67107-9b7b-40f3-b09a-d8efa9d92766.mp4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,7 @@ qt5_wrap_cpp(PLOTJUGGLER_BASE_MOCS
   plotjuggler_base/include/PlotJuggler/transform_function.h
   plotjuggler_base/include/PlotJuggler/plotwidget_base.h
   plotjuggler_base/include/PlotJuggler/lua_highlighter.h
+  plotjuggler_base/include/PlotJuggler/script_editor.h
 )
 
 add_library( plotjuggler_base
@@ -170,6 +171,7 @@ add_library( plotjuggler_base
      plotjuggler_base/src/timeseries_qwt.cpp
      plotjuggler_base/src/reactive_function.cpp
      plotjuggler_base/src/lua_highlighter.cpp
+     plotjuggler_base/src/script_editor.cpp
      )
 # target_link_libraries(plotjuggler_base plotjuggler_qwt)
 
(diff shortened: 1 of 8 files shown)


facontidavide:
This looks like a great addition, but I would prefer not to add the **Qt_License_Info** in the **Help** menu.

Showing that license and not the LGPL license of PlotJuggler itself could be misleading.

Would you please remove that from this PR?

facontidavide (changes requested):
please remove the Qt license dialog from Help menu

Bartimaeus-:
> This looks like a great addition, but I would prefer not to add the **Qt_License_Info** in the **Help** menu.
> 
> Showing that license and not the LGPL license of PlotJuggler itself could be misleading.
> 
> Would you please remove that from this PR?

Hi Davide, thanks for the feedback. I did wonder if you would be okay with the license info addition. The widget is part of the Qt toolbox and is licensed under BSD with the stipulation that "Redistributions in binary form must reproduce the above copyright  notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution."

I also identified this library which is under MIT and has its own syntax highlighting with Lua support built-in and a completer, but would add more source files than the Qt toolbox widget: https://github.com/Megaxela/QCodeEditor

The options I see that avoid license issues:

- Leave the Qt license info dialog in Plotjuggler, with possible relocation to remove it from the Help menu and/or addition of the Plotjuggler project license. Other text could also be added as necessary to add clarity for the licenses (ex: "Plotjuggler uses components licensed under the following terms:")
- Use another widget such as the QCodeEditor linked above, with possibly larger code footprint and replacement of existing Lua highlighter
- Write a widget from scratch (I don't have a very big appetite for this)

Let me know what you prefer. I started with this as what I thought would be the simplest to execute, but am open to working on this some more if needed.

facontidavide:
I think I will go all-in with https://github.com/Megaxela/QCodeEditor (that was already on my radar)

facontidavide:
Good news. I integrate QCodeEditor :) 

We also gain autocomplete and autoindent 

image

PR #666, opened 2022-06-15
Fix CSV generated time axis.

agoessling:
Previously the CSV dataload plugin was not saving the correct XML state
when a generated time axis was used.


agoessling:
@facontidavide Any thoughts on this?

facontidavide:
thanks

PR #673, opened 2022-06-24
Added support for converted int types

RanoVeder:
Currently the parquet loader plugin only supports INT32 and INT64 parquet types. This commit adds supports for the following extra data types by using the ConvertedType from each column:

UINT8
UINT16
UINT32
UINT64
INT8
INT16


facontidavide:
thanks! I was not sure if anyone was interested to this plugin!

PR #679, opened 2022-07-01
Add line numbers to csv loader

Bartimaeus-:
Promote the rawText widget in csv loader to QCodeEditor for line number display and set the widget theme to be consistent with the Lua script editor (no syntax highlighting)

image


facontidavide:
is it still read-only?

Bartimaeus-:
> is it still read-only?

Yes
image

PR #680, opened 2022-07-02
Add CSV table preview and CSV highlighting

Bartimaeus-:
**CSV Table Preview**
-CSV data loader updated with new table view widget for file preview
-Table view updates when user changes delimiter
-Tab widget added to the CSV data loader to switch between raw text view and table view

https://user-images.githubusercontent.com/2954254/177014679-7dbf53e0-b374-4d22-adb5-0862a1d23c37.mp4

image

**CSV Highlighting**
-Add QCSVHighlighter to apply syntax highlighting to the selected delimiter in the csv raw text preview
-Highlighting updates when user changes delimiter

image

### DIFF
--- a/3rdparty/QCodeEditor/CMakeLists.txt
+++ b/3rdparty/QCodeEditor/CMakeLists.txt
@@ -28,6 +28,7 @@ set(INCLUDE_FILES
     include/internal/QHighlightBlockRule.hpp
     include/internal/QCodeEditor.hpp
     include/internal/QCXXHighlighter.hpp
+    include/internal/QCSVHighlighter.hpp
     include/internal/QLineNumberArea.hpp
     include/internal/QStyleSyntaxHighlighter.hpp
     include/internal/QSyntaxStyle.hpp
@@ -47,6 +48,7 @@ set(SOURCE_FILES
     src/internal/QCodeEditor.cpp
     src/internal/QLineNumberArea.cpp
     src/internal/QCXXHighlighter.cpp
+    src/internal/QCSVHighlighter.cpp
     src/internal/QSyntaxStyle.cpp
     src/internal/QStyleSyntaxHighlighter.cpp
     src/internal/QGLSLCompleter.cpp
--- a/3rdparty/QCodeEditor/include/QCSVHighlighter
+++ b/3rdparty/QCodeEditor/include/QCSVHighlighter
@@ -1,0 +1,3 @@
+#pragma once
+
+#include <internal/QCSVHighlighter.hpp>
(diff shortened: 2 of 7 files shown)


Bartimaeus-:
Added an update to have checkboxes for enabling these new features. These are disabled by default so the original behavior and performance of the CSVloader is retained.

image

image

image

facontidavide:
Thnaks a lot. I will remove the enable button. 100 rows is not a problem

Bartimaeus-:
I found when stress testing it that performance takes a hit as columns
increases. Syntax highlighting is still pretty responsive but with 100
columns for example there's a 2-3 second delay with the table view. That
would start to add up if you needed to open multiple files with large
numbers of fields, even if they aren't very large files.

Bartimaeus-:
The delay occurs when the table is built, so initial loading and changing
the delimiter are where the performance hit occurs

facontidavide:
In that case, I will reduce the number of lines to 50 and call it a day.

Can you give me one of those CSV files that you use for stress test?

image

Bartimaeus-:
Here is an example file plus python script to generate csv files with fake data
[csv_generator.zip](https://github.com/facontidavide/PlotJuggler/files/9079382/csv_generator.zip)

facontidavide:
Please try the latest commit https://github.com/facontidavide/PlotJuggler/commit/0ff91357acf945e02138bc47755b5b39f7bc3044

On my computer it is pretty fast, even with your test CSV and compiled in Debug mode.

facontidavide:
By the way: 
image

Bartimaeus-:
> Please try the latest commit [0ff9135](https://github.com/facontidavide/PlotJuggler/commit/0ff91357acf945e02138bc47755b5b39f7bc3044)
> 
> On my computer it is pretty fast, even with your test CSV and compiled in Debug mode.

Just ran it, I think it's tolerable for a single file. The loading times are more noticeable imo if you make 2-3 copies of the csv and try to open them all at once. My screen recorder seems to jump past the loading times in the captured video (the mouse pointer even jumps from one spot to another on the screen) so it's not easy to show a side-by-side against the original loader.

I think an upper limit on the columns displayed would be a good idea, even if that number is large. I'd hate for the preview to be the bottleneck for someone's use case, the rest of PJ seems to rip through large datasets with ease. The most likely use case that comes to mind is using PJ to combine multiple input files into larger CSVs and then loading those larger csvs.

I don't think it's worth the effort but a fancy way to cap things would be to cap the total number of cells that are added to the table, so as columns increases rows decrease. Capping the column count seems just as effective and a lot simpler though.

Bartimaeus-:
This is definitely fun to work on, and I'm stoked to try out the new features you just added!

PR #682, opened 2022-07-09
Fix start/end time bug in CSV Exporter

Bartimaeus-:
Fix bug with the csv publisher where clicking buttonGetStart or buttonGetEnd buttons would override the checkBoxFirst/checkBoxLast selections

image

~~Fixed by enabling/disabling the buttons when the checkboxes are toggled~~
**Edit:** fixed by unchecking the first/last checkboxes when the button is pressed. This seems less of an annoyance where a user may move the cursor and want to set the time and not notice that the first/last checkbox is checked


Bartimaeus-:
Changed the behavior so that the buttons remain clickable and they will uncheck the first/last checkboxes when clicked. This should avoid the potential annoyance of having to uncheck the first/last checkboxes when trying to manually set the start and end times.

PR #700, opened 2022-07-23
Fix minor include problem for mqtt_plugin

ozzdemir:
In mqtt_plugin, std::thread is being utilized, but instead QThread is being included. This results in include problems on my setup (Ubuntu 18.04) on source builds (an sample error in below). The minor change proposed resolves the include problems in my setup.

```
PlotJuggler/plotjuggler_plugins/DataStreamMQTT/datastream_mqtt.h:53:8: error: ‘thread’ in namespace ‘std’ does not name a type
   std::thread _mqtt_thread;
```


facontidavide:
thanks

PR #702, opened 2022-07-25
Statistics dialog improvements and bug fixes

Bartimaeus-:
Changes and bug fixes to the statistics dialog window

Improvements:
- Statistics can now be calculated from the visible range or by the full data range (previously only visible range)
image

- Statistics window title now includes the title from the plot
- Made the dialog window stay on top of the PlotJuggler window but not on top of other programs. The statistics dialogs will minimize/restore with the main PJ window. Previously the statistics dialog windows would stay on top of all windows

- ~~Made the table sortable by column (click on the header item to sort by that column)~~  (**Edit:** disabled this as there are bugs. earlier commit kept as it set the styling for the arrows)

Bug fixes:
- Fix bug where statistics would only get updated for the plot with the mouse over it when zooming while plot zooms were linked
- Fix bug where the statistics weren't updated when the zoom reset button was pressed
- Fix bug where series weren't removed from the statistics dialog right away after a curve was removed from the plot

### DIFF
--- a/plotjuggler_app/plot_docker.cpp
+++ b/plotjuggler_app/plot_docker.cpp
@@ -306,6 +306,7 @@ DockWidget::DockWidget(PlotDataMapRef& datamap, QWidget* parent)
 
   connect(_plot_widget, &PlotWidget::splitVertical, this, &DockWidget::splitVertical);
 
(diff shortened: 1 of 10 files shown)


Bartimaeus-:
I'm not able to test the "full range" mode with a streaming timeseries (tested with csv files). I'm not sure if it will update as new data points are added to the series. Help with testing this would be appreciated

PR #714, opened 2022-08-04
Fix/snap rosbag

Guillaumebeuzeboc:
This PR slightly alters the snap, resulting in the following user experience,
Install the snap,

`sudo snap install plutjuggler`

Launch the snap with ROS support,

`plotjuggler`

This is also achieved hitting the desktop launcher. It is on par with the behaviour of the current ‘stable’ snap. So far, no disruption for users.

The actual change is the introduction of two new commands,

`plotjuggler.ros`
`plotjuggler.ros2`

Which respectively (explicitly) launch PJ with ROS or ROS 2 support.
Finally, users are able to change the behaviour of the ‘plotjuggler’ command as follows,

`sudo snap set plotjuggler ros-plugin-version=2`

So that the command now launches PJ with ROS 2 support instead of ROS.

This change is the result of fixing the issue with loading rosbags.

Being able to load rosbag brought up two main problems:
- Due to libraries that have the same name (`libclass_loader.so`) both ROS/ROS 2 plugins cannot be loaded at the same time
- ROS 2 bag `metadata.yaml` are referring to the `.db3` file relatively. In snaps GUI with Qt, files loaded with `QFileDialog` are fuse-mounted to `/run/user/<uid>/doc/<hash>/` in order to give your snapped application access to it (from the[ xdg-desktop-portals](https://snapcraft.io/docs/xdg-desktop-portals)). The `.db3` file is then searched within `/run/user/<uid>/doc/<hash>/` which only contains the `metadata.yaml`.

This PR also includes:

- GitHub workflow action is now checking out the full history of the repository since we need the latest tag for the snap versioning.
- ZMQ build was removed since now ZMQ can be found from the deb package

This PR does not disrupt any existing install while bringing ROS 2 support along with an explicit command.

Feel free to ask any questions regarding the snap or its usage.

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -17,6 +17,9 @@ jobs:
       snap-file: ${{ steps.build-snap.outputs.snap }}
     steps:
     - uses: actions/checkout@v2
+      with:
+        # full history for latest tag name
+        fetch-depth: 0
     - uses: snapcore/action-build@v1
       id: build-snap
 
--- a/snap/gui/PlotJuggler-ros2.desktop
+++ b/snap/gui/PlotJuggler-ros2.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=PlotJuggler-ROS2
+Exec=plotjuggler.plotjuggler-ros2
+Icon=${SNAP}/meta/gui/plotjuggler.svg
+Comment=Visualize timeseries like a pro
+Terminal=false
+Categories=Development;
+Name[en_US]=PlotJuggler-ROS2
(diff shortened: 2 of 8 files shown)


facontidavide:
since I don't understand any of this... I will just merge it. Thanks!

facontidavide:
I just released the tag 3.5.2, but it looks like it is still in devmode:

image

Guillaumebeuzeboc:
The CI released the snap under the "stable" grade : https://github.com/facontidavide/PlotJuggler/runs/7689122227?check_suite_focus=true#step:3:3337

And the was published to the candidate channel: https://github.com/facontidavide/PlotJuggler/runs/7689877367?check_suite_focus=true#step:3:201

So everything went well on the CI side.

The [plotjuggler snapstore page](https://snapcraft.io/plotjuggler) doesn't show this `candidate` release, and this is not normal.
I don't why yet. I will ask internally and come back to you once I know more

Issue #716, opened 2022-08-05
Some Mouse Events not Captured After Drag and Drop


Bartimaeus-:
**Edit:** some of the content here incorrectly describes the behavior of what was happening. See https://github.com/facontidavide/PlotJuggler/pull/717

**Summary**
Some mouse events are not handled by most plot widgets after one or more curves are added to an empty plot widget

**Setup**
-Open a new plot widget (Plot 2)
-Add curve(s) to Plot 2

**Issue**
-After curve added to Plot 2, some mouse events (right/left/middle-click) are not handled by Plot 1 (mouse scroll wheel events are still handled ). Only the plot widget itself seems to have this issue-- the plot toolbar and plot legend are not impacted

Mouse event handling in the other open plots is restored after any of the following occurs:
-A mouse event occurs in any other part of PlotJuggler besides Plot 1's plotting area. This includes Plot 2 (which is handled)
-The zoom changes in Plot 1 using a mouse scroll wheel event
-The mouse moves over any of the green shaded region indicated in the figure below

![image](https://user-images.githubusercontent.com/2954254/183125082-feca408f-3391-4576-9c63-8e7103c8aa8f.png)

Other Notes:
-Only verified on Windows (Linux not checked)
-Behavior is present at least back to 3.5.0 release (earlier versions not checked)

Bartimaeus-:
I've figured out how to fix most of this by clearing the drag info and curves after a mouse button release event:

(snippet from canvasEventFilter() in plotwidget.cpp)
```
    case QEvent::MouseButtonRelease: {
      if (_dragging.mode == DragInfo::NONE)
      {
        QApplication::restoreOverrideCursor();
        return false;
      }
      _dragging.mode = DragInfo::NONE;     // new line added
      _dragging.curves.clear();    // new line added
    }
```

While testing this solution I found this fixes everything except for the middle mouse pan, but strangely this is only applicable to the first plotwidget. This is somehow due to _dragging.mode not getting reset to DrafInfo::NONE

To demonstrate the issue, I've added the following code:

```
    case QEvent::MouseButtonPress: {
      if (_dragging.mode != DragInfo::NONE)
      {
          QMessageBox::warning(qwtPlot(), "Drag in process",
                               tr("Mouse button press event but there's a drag in process!"));
          return true;  // don't pass to canvas().
      }
```

[...]

PR #717, opened 2022-08-06
Fix plot widget drag and drop bug (Issue #716)

Bartimaeus-:
- Plot widgets were not clearing their drag and drop state if a drag event left the widget, causing mouse events to not be handled properly if the drop event happened in another plot widget
- Added event handlers for drag leave event

(This was actually a pain to debug at first because it seemed like it was only impacting the first plot widget that existed, but it was just unfortunate that I was really repeatably dragging across the same widget)


facontidavide:
fabtastic, thanks!

PR #731, opened 2022-08-21
Add CSV loader date-time format help dialog

Bartimaeus-:
Added a new dialog window to the CSV loader to help with the date time format strings.

image

**Features:**
-Date time format string preview: displays the result from a given format string for a user-specified date time
-Checkbox at the bottom to allow the help dialog to stay open when opening multiple csv files

The dialog can be opened by clicking on the new "Date Format Help" button that replaces the old message with hyperlink to the Qt documentation


facontidavide:
Thanks for contributing. I will be away from the computer for the next 2 weeks. I will review this PR afterward

PR #736, opened 2022-08-25
Adds a websocket client

mtan1503:
Added a websocket client to the options of data streamers. Sometimes it is not possible to use a websocket server because a server already exists. To connect to the websocket, the user enters a URL in the textbox so that data from any endpoint can be displayed.

image


facontidavide:
Instead of adding a new plugin, I think that it would be better to extend the existing one. What do you think?
Maybe a [QRadioButton](https://doc.qt.io/qt-6/qradiobutton.html) to select either client or server

mtan1503:
Yes, that's a good idea. We are now working with a fork of this repository so for now that serves our purpose. If you wish to integrate the two functionalities, be my guest!

facontidavide:
I

facontidavide:
stale

PR #726, opened 2022-08-17
Add Drag n Drop

bonkuraps:
- Add drag n drop functionality
- Check if manually provided plugin name is enabled
- Update `.gitignore`
- Run `clang-format`

And thanks for inspiration: https://github.com/facontidavide/PlotJuggler/pull/454

PR #739, opened 2022-09-02
Fix multi-plugin selection

zdavkeos:
Broken in #726. If all plugins are enabled, then opening a file supported by multiple plugins does not work.


facontidavide:
Thanks

PR #745, opened 2022-09-19
Set MQTT topics list to be multi-selectable

avanover:
PlotJuggler has become an integral part of our workflow for testing our system. Previous version of PJ allowed you to subscribe to multiple MQTT topics, but since the change to the mosquito-based MQTT plugin, this is no longer possible. This change fixes that and allows us to plot data from multiple topics once more.


avanover:
Addresses #747

facontidavide:
thanks

PR #746, opened 2022-09-19
Always use topic names for creating MQTT parsers

avanover:
The MQTT data streamer in plot juggler currently does not use the MQTT topic name when creating parsers. If/when #745 is merged, this becomes a larger issue as data from multiple different topics becomes merged together. If different MQTT topics contain fields of the same name, then it is next to impossible to differentiate them in PJ. This change allows the topic name to be reflected in the tree view when selecting data and makes the data fields easily distinguishable.


facontidavide:
thanks

PR #751, opened 2022-09-21
Add a new meme with The Rock

cosmicog:
plot_with_the_rock


cosmicog:
Max meme file size was 83KB, this is 261KB, should I clean the history and re-create pr with an optimized version?

junwoo091400:
Nice one!

Issue #752, opened 2022-09-23
ros2 Humble snap 3.6: cannot launch plotjuggler with Cyclone DDS


doisyg:
# Problem description
With the last snap version (3.6), starting plotjuggler from a terminal which has its env variable `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp `blocks the launch of plotjuggler

> [ERROR] [1663971926.410750860] [rcl]: Error getting RMW implementation identifier / RMW implementation not installed (expected identifier of 'rmw_cyclonedds_cpp'), with error message 'failed to find shared library 'rmw_cyclonedds_cpp', at /tmp/binarydeb/ros-foxy-rmw-implementation-1.0.3/src/functions.cpp:73', exiting with 1., at /tmp/binarydeb/ros-foxy-rcl-1.1.13/src/rcl/rmw_implementation_identifier_check.c:139

# Steps to reproduce

Jammy / ROS2 Humble

> $ export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
> $ plotjuggler.ros2

mayak1fm:
A cyclone is installed  on mu PC, but it stiil crashes with error:
` Error getting RMW implementation identifier / RMW implementation not installed (expected identifier of 'rmw_cyclonedds_cpp'), with error message 'failed to find shared library 'rmw_cyclonedds_cpp', at /tmp/binarydeb/ros-foxy-rmw-implementation-1.0.3/src/functions.cpp:73', exiting with 1., at /tmp/binarydeb/ros-foxy-rcl-1.1.13/src/rcl/rmw_implementation_identifier_check.c:139`

Guillaumebeuzeboc:
Regarding the missing `cyclonedds` implementation, the current snap doesn't contain the `ros-foxy-rmw-cyclonedds-cpp` package.
It must be added as a `stage-package` dependency in order for it to be distributed with the snap.
I have made the change [here](https://github.com/Guillaumebeuzeboc/PlotJuggler/tree/fix/cyclonedds).
You can test it by downloading the `.snap` file available as artefact [here](https://github.com/Guillaumebeuzeboc/PlotJuggler/actions/runs/3244585435) (from my simple test it seems to work).

[...]

PR #763, opened 2022-11-10
Fix/add other dds vendors

Guillaumebeuzeboc:
Add additional DDS vendors to the snap.

This will fix https://github.com/facontidavide/PlotJuggler/issues/752

Since the snap is isolated, additional DDS vendors must be added to the snap itself.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -287,6 +287,8 @@ parts:
       - ros-foxy-vrxperience-msgs
       - ros-foxy-webots-ros2-msgs
       - ros-foxy-wiimote-msgs
+      # Cyclonedds
+      - ros-foxy-rmw-cyclonedds-cpp
     override-pull: |
       if [ ! -d plotjuggler-ros-plugins ]; then
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -287,8 +287,10 @@ parts:
       - ros-foxy-vrxperience-msgs
       - ros-foxy-webots-ros2-msgs
       - ros-foxy-wiimote-msgs
-      # Cyclonedds
+      # Additional DDS vendors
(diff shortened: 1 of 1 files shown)


facontidavide:
question: how will ROS2 inside the snap select a certain vendor over the others?

Guillaumebeuzeboc:
> question: how will ROS2 inside the snap select a certain vendor over the others?

I checked using the environment variable (`RMW_IMPLEMENTATION`) as described in [this documentation](https://docs.ros.org/en/foxy/How-To-Guides/Working-with-multiple-RMW-implementations.html).

Now that you are asking, I wonder if it would be nicer to offer the possibility to the user to set the DDS vendor for a long term.

An example would be
```
sudo snap set plotjuggler ros-rmw-implementation=rmw_connext_cpp
```
And then Plotjuggler ROS 2 would automatically use the selected vendor.

What do you think?
@artivis, any thoughts about that?

artivis:
I would avoid introducing a PJ snap specific way of setting this parameter as this can potentially introduce confusion. E.g. what would take precedence between the snap parameter and the environment variable if both are set to different value?

Guillaumebeuzeboc:
> I would avoid introducing a PJ snap specific way of setting this parameter as this can potentially introduce confusion. E.g. what would take precedence between the snap parameter and the environment variable if both are set to different value?

In this case, it would be the environment variable set (the parameter would be for people would always use a specific vendor and simply want to launch plotjuggler from the desktop icon).

I don't know if it would be useful to anyone, so maybe for now we can leave it with simply the environment variable.

artivis:
Is this PR blocked by anything? Looks like it is fixing the initial issue and it's been tested by @doisyg as reported in #752

facontidavide:
blocked by my little time to think about Plot Juggler. Thanks!!

artivis:
Eheh sorry Davide :D 
Right back at you!

PR #767, opened 2022-12-13
[bugfix] Allow for NaN floating point numbers

markcutler:
I have been playing around with the mcap data storage type and noticed that messages with NaN entries throw a truncation error in the latest PlotJuggler release.

Here we first check for NaN before doing the truncation check.

Please let me know if there are unit tests I can add/update. Thanks.


facontidavide:
Thanks a lot for finding this issue.

I would probably move this code somewhere else. Let me think about it

facontidavide:
implemented more broadly. Thanks

PR #768, opened 2022-12-13
[bugfix] Allow for very large uint64_t types

markcutler:
Again exploring mcap data storage type (not sure if this bug also exists with other types):

Very large uint64_t types fail this check and will not load in the latest PlotJuggler: 
```
if( from != static_cast<SRC>(static_cast<DST>( from)))
        throw RangeException("Floating point truncated");
```

For instance, 
```
    uint64_t val_int = std::numeric_limits<uint64_t>::max() - 1000;
    double val_double = 0;
    val_double = static_cast<double>(val_int);
    uint64_t val_int2 = static_cast<uint64_t>(val_double);

    std::cout.precision(25);
    std::cout << "val_i1: " << val_int << std::endl;
    std::cout << "val_d1: " << val_double << std::endl;
    std::cout << "val_i2: " << val_int2 << std::endl;
```
returns
```
val_i1: 18446744073709550615
val_d1: 18446744073709551616
val_i2: 0
```
presumably because the `double` does not have enough precision to accurately represent the large `uint64_t`.

The proposal here could be a bit of a hack, I'm not sure. However, it solves the problem on my machine and is likely supported on most modern architectures. The idea is that all uint64_t values can be accurately represented by a larger data type so use a `double long` during the conversion/checking process, and then cast back to `double` when you are confident that the data are valid.

Let me know if there is a better fix.

### DIFF
--- a/plotjuggler_plugins/ParserROS/rosx_introspection/include/rosx_introspection/variant.hpp
+++ b/plotjuggler_plugins/ParserROS/rosx_introspection/include/rosx_introspection/variant.hpp
@@ -304,27 +304,27 @@ template<typename DST> inline DST Variant::convert() const
 template<> inline double Variant::convert() const
 {
   using namespace RosMsgParser::details;
-  double target = 0;
+  double long target = 0;
   const auto& raw_data = &_storage.raw_data[0];
   //----------
   switch( _type )
   {
   case CHAR:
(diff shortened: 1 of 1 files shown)


facontidavide:
The problem I have with this is that the truncation will happen anyway, just at some other point in the code!

This change doesn't really fix anything. The only real solution is to accept truncation

PR #770, opened 2022-12-13
[bugfix] conversion typo?

markcutler:
Fixing what I assume is a typo in the code that I noticed while looking at other things. I didn't see `checkUpperLimitFloat` used anywhere else so I assume it was meant to be here.


facontidavide:
thanks

PR #771, opened 2022-12-22
Update Reactive Scripts on playback loop

augustinmanecy:
update the reactive scripts on playback loop. previously on moving manually the timeslider updated them.

PS: thanks for this amazing tool, I use it everyday!


facontidavide:
thanks

Issue #201, opened 2019-09-12
Command line flags instead of pop-up windows on startup from layout file


lucasw:
I'm starting plotjuggler from a launch file with a layout file and I'd like it to jump right into plotting my ros topics without having to click past anything.

```
Start the previously used streaming plugin?

 ROS_Topic_Streamer 
```
Want to be able to skip past this by automatically entering 'yes'.

```
Select ROS messages
...
```
Skip this with equivalent  'OK' getting entered.

```
One or more timeseries in the layout haven't been loaded yet
What do you want to do?
```

Automatically create empty curves- removing the curves seems contradictory to loading the layout in the first place.

facontidavide:
I will think about it... I don't want this to become a source of hard to debug errors....

ingramator:
Definitely agree with @lucasw, it's very annoying having to use mouse input every time you launch. Was this issue ever addressed? I'd be happy to look into doing a PR.

normanhendrich:
I also need this. When launching plotjuggler with an existing layout/configuration xml, it should come up quietly together with all other launched nodes, without needing these distracting mouse clicks.

valschmidt:
I also would appreciate this feature.

drdelayismysenpai:
This feature is so basic yet so crucial for many launch files.

I would also appreciate it.

facontidavide:
Relevant link: https://github.com/facontidavide/PlotJuggler/discussions/520

ljburtz:
Hello and thank you for plotjuggler's amazingness

I understand features take lots of time to implement, just adding to the voices of users above (to nudge towards -> "It has a considerable impact on a large percentage of my users.")

As a ros user, the feature to save the input topics in the layout file would:
- avoid clicks (as above)
- avoid forgetting which topics are required (always getting the 'missing timeseries' error message when too many topics to just 'select all')
- big use case for automation -> make it possible to launch plotjuggler in roslaunch, the same way that rviz can be roslaunched with it's config/layout and visualizations immediately start updating.

Thanks a lot for your consideration

awesomebytes:
Hello, I attempted to add the functionality to automate starting PlotJuggler skipping dialogs that need human clicks.
I believe it's unlikely to be merged given I used environment variables to allow for this, however, if building from source is okay for you, it is a possibility now!

The environment variables introduced are:
[...]

PR #811, opened 2023-04-02
Introduce environment variables to skip dialogs to automate PlotJuggler usage

awesomebytes:
First, @facontidavide let me apologize for doing this PR as I am the first one to see it is an ugly workaround, however, the functionality I think is great to have so...

Introducing the environment variables:
`PLOTJUGGLER_ACCEPT_PREVIOUSLY_USED_STREAMING_PLUGIN=1` to skip the dialog 'Start the previously used streaming plugin?' 
`PLOTJUGGLER_CHOOSE_EMPTY_PLACEHOLDERS_ON_MISSING_TIMESERIES=1` to skip the dialog 'One or more timeseries in the layout haven't been loaded yet, What do you want to do?'

Note that there is a sibling commit in plotjuggler-ros-plugins (https://github.com/PlotJuggler/plotjuggler-ros-plugins/pull/64) to skip the ROS topic selection dialog.
This is a workaround for https://github.com/facontidavide/PlotJuggler/issues/201 (passing commandline flags would require a lot of changes).

P.S.: I would have loved to pass it via commandline, but it required a lot of changes. And the plugin side would have needed many more to make it elegant. Maybe `PLOTJUGGLER_ACCEPT_PREVIOUSLY_USED_STREAMING_PLUGIN` could be implemented as a commandline plugin. Which, by the way, I did not understand what `--start_streamer` does... I thought it would do that, but I could make it do anything. My bad.
P.S.2: I do not expect this to be merged due to the lack of elegance, specially being the main PlotJuggler repo. But maybe people will want it enough to build from source. (I may).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,14 @@ Note that this also affect the desktop launcher.
 
 You can find find the detailed instructions here: [COMPILE.md](COMPILE.md).
 
+## Auto start previously used streaming plugin
+When loading a layout that includes a streaming plugin, PlotJuggler asks `Start the previously used streaming plugin?`, in order to automatically
+click on `Yes`, you can set the environment variable: `PLOTJUGGLER_ACCEPT_PREVIOUSLY_USED_STREAMING_PLUGIN=1` and this dialog will be skipped.
+
+## Automatically choose 'Create empty placeholders' when data from a layout is missing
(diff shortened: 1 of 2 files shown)


facontidavide:
Isn't it better to add a command line argument?

facontidavide:
wouldn't be easier to add this in Preferences?
In particular, PLOTJUGGLER_CHOOSE_EMPTY_PLACEHOLDERS_ON_MISSING_TIMESERIES, should be a Preference

facontidavide:
also, Windows CI is unhappy

facontidavide:
close for inactivity

PR #824, opened 2023-05-14
Handle protobuf maps

adobke:
Hi,

Currently protobuf maps are treated as a list of repeated messages where the name of the series is the index in the list of repeated messages. Unfortunately, the order of the entries in the map is not stable, and could change over time -- as a result, we end up with a single series effectively containing a random set of map entry, which is not so useful for plotting!

This change detect protobuf maps and uses the "key" in the series name instead of just the index allowing us to plot protobuf map data!

I added an example of this to the MQTT publisher and used that to test this change.

I'm happy to make any other changes to help add this feature, thank you!


facontidavide:
looks good. thanks a lot!

Issue #832, opened 2023-05-25
Add support for custom timestamps for real-time streaming using Protobuf over UDP


rinnaz:
Thanks for your great work!

### Problem description

I'm using PJ 3.7.1 installed via snap and I'm having an issue with real-time streaming using UDP with Protobuf  serializer. 
The thing is that PJ is not handling "timestamp" field in Protobuf message and is using timestamp that is generated on message receive event. This leads to visible jitter.

On the left side you can see default plot and on the right side XY plot using message timestamps.
![image](https://github.com/facontidavide/PlotJuggler/assets/36293513/50a2fbcd-b266-4aa9-b745-4c45ecd46987)

PR #835, opened 2023-05-26
Add support for Protobuf message timestamp

rinnaz:
Please see #832 for description.

Now Protobuf Parser uses value from message "timestamp" field if such is available.


facontidavide:
just pushed a cherry-picked change (added a checkbox to enable this feature).


thanks a lot and let me know if this works for you

PR #842, opened 2023-06-14
Showing timestamp off of beginning of log

jbendes:


facontidavide:
not a good idea. It is recommended to use time since epoch by default

PR #843, opened 2023-06-16
Add link to QDate format string

faisal-shah:
A link to QTime format string was there, but not QDate


Bartimaeus-:
I made a pop-up help dialog for the DataLoaderCSV that shows the formatting info for QDate and QTime. The dialog is in its own set of files and it should be easy to let other plug-ins use it (would also be more unified experience)

https://github.com/facontidavide/PlotJuggler/pull/731

Issue #828, opened 2023-05-23
no option for tab delimiter


aaaddd4747:
# Problem description

i use tab characters in my CSVs (as is [accepted in the CSV spec](https://csv-spec.org/)) because it produces much nicer results when printed to a terminal.
PlotJuggler does not support this. 

# Steps to reproduce (important)

load up PlotJuggler,
go to load in your CSV,
go to change to the correct delimiter,
be sad when the right one isnt in the list.

# Suggested solution

feature request: add tab delimiters as a feature

PR #844, opened 2023-06-17
Add 'tab' as a separator in the CSV loader

zdavkeos:
Addresses #828

tab_option_csv


facontidavide:
thanks for contributing

PR #846, opened 2023-06-21
Use a more reliable method to select buildtool

paulbovbel:
The current approach doesn't work in our build system, since you need `find_package(catkin QUIET)` to fire. The other env vars aren't set in our build. I believe this method should be a little more bulletproof, and works fine on our build of https://github.com/BehaviorTree/BehaviorTree.CPP.


paulbovbel:
Ping on this? we've been using it in our ROS1 and ROS2 internal builds for ages, and would love to de-fork

PR #849, opened 2023-06-27
update to core22, remove ros1, enable humble instead of foxy

MirkoFerrati:
Given that the newer ROS versions require ubuntu22 as the base OS, an update to the snapcraft files is needed and this PR brings the snap to core22. 
Since ROS1 is no longer supported in 22, I had to duplicate the pipelines in the old ros1/ros2/core20 and the new ros2/core22. 
The existing snap is still the default, if a user needs the new snap with core22 they can install by manually changing the snap track to humble in their machine. 
Eventually the new default should become core22 and at that point users still on ros1 will have to change their snap track.
Requires https://github.com/facontidavide/PlotJuggler/pull/853 before CI can pass.

PR #867, opened 2023-09-15
Remove deprecated messages from foxy

MirkoFerrati:
It seems astuff/pacmod msgs are changing frequently and the original name is no longer present in the ppa


zdavkeos:
This would allow some of the open pr's to pass ci

MirkoFerrati:
> This would allow some of the open pr's to pass ci

@zdavkeos 
This commit is already included in https://github.com/facontidavide/PlotJuggler/pull/849 which is going to be merged soon (I think)

Issue #866, opened 2023-09-11
Plotting values without connecting them with a line but as a step


marios-stam:
Hello all and thank you for this amazing software.

i was wondering if there is feature or somehow can be implemented to plot data without connecting 2 successive points with just a line   but have a step effect on the plotting. Something similar to the below:
![image](https://github.com/facontidavide/PlotJuggler/assets/37310507/60baeb05-ed8f-40ef-b17b-d8213f154d38)
 
Right now the plot is like the left and want to make it like the right.

Thank you very much!

marios-stam:
I am sorry for not mentioning it, although I think you will notice either way, that these specific plots were made through pyplot just to showcase what I am trying to say.

zdavkeos:
This can be closed since #869

PR #869, opened 2023-09-20
Add "Steps" when drawing curves

zdavkeos:
As requested by issue #866. Step-wise interpolation between points (already supported by qwt).

pj_steps


facontidavide:
Thanks for contributing

PR #872, opened 2023-10-07
Remove Steven Crowder meme

AlexanderDavid:
Steven has become much more problematic since this meme was added.


facontidavide:
I don't know who steven crowder is, nor why he is "problematic".

I have my personal opinions, but I don't express them through my memes

Issue #879, opened 2023-10-27
Line style post step


mikkelti:
# Feature request
I love the new `Steps` line style, but I would also like to see "post steps", where the line has the value of `y[i]` on the interval `(x[i], x[i+1])`.

PR #880, opened 2023-11-02
#879 Inverted option for STEPS style

zdavkeos:
New line style like STEPS but drawn the opposite way.
Screenshot 2023-11-01 at 7 16 40 PM


zdavkeos:
@facontidavide & @mikkelti I am not crazy about the name "Inverted Steps", that's qwt's terminology. I would be very open to any suggestions.

mikkelti:
@zdavkeos Inspired by matplotlib's terminology, you could call them `Steps (pre)` and `Steps (post)` or something like that.

plotjuggler_app/plotwidget_editor.ui
```diff
@@ -344,7 +344,14 @@
...
               <property name="text">
-               <string>Steps</string>
+               <string>Steps (pre)</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <widget class="QRadioButton" name="radioStepsInv">
+              <property name="text">
+               <string>Inverted (post)</string>
```
facontidavide, line 354:
```suggestion
               <string>Step (post)</string>
```

facontidavide (changes requested):

zdavkeos:
Good catch @facontidavide , Thanks!

PR #884, opened 2023-11-22
Allow multiple publishers to ZMQ Data Stream

jmount-boss:
I couldn't see a way for PlotJuggler to subscribe to multiple ZMQ publishers. However, if the PJ ZMQ Data Stream subscriber changes `connect` to `bind` then multiple publishers can use the same endpoint and PJ gets all the data (Stack Overflow question for [reference](https://stackoverflow.com/a/67752410)). 

This is feature is useful in a few scenarios. For example,
- when you have more than one application running and want to visualise data from both at the same time; or
- when you wish to run the same algorithm but with slightly different parameters and visualise the difference in real-time.

This implementation would require existing PlotJuggler users employing the ZMQ DataStream plugin to change their publisher from bind to connect in their code.

Can be tested using the following two Python Scripts

```python
# pub1.py
import math
import time
import zmq

if __name__ == "__main__":
    context = zmq.Context()
    publisher = context.socket(zmq.PUB)
    publisher.connect("tcp://127.0.0.1:12345") 

    idx = 0
    while True:
        time.sleep(1)
        val = math.sin(idx)
        data = {"idx1": idx, "val1": val}
        publisher.send_json(data)
        idx += 1
        print("Published:", data)
```

```python
# pub2.py
import math
import time
import zmq

if __name__ == "__main__":
    context = zmq.Context()
    publisher = context.socket(zmq.PUB)
    publisher.connect("tcp://127.0.0.1:12345") 

    idx = 0
    while True:
        time.sleep(1)
        val = math.cos(idx)
        data = {"idx2": idx, "val2": val}
        publisher.send_json(data)
        idx += 1
        print("Published:", data)
```

### DIFF
--- a/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
+++ b/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
@@ -126,7 +126,7 @@ bool DataStreamZMQ::start(QStringList*)
       (dialog->ui->comboBox->currentText() + address + ":" + QString::number(port))
           .toStdString();
 
-  _zmq_socket.connect(_socket_address.c_str());
+  _zmq_socket.bind(_socket_address.c_str());
 
   parseTopicFilters(topics);
   subscribeTopics();
--- a/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
+++ b/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
@@ -126,7 +126,14 @@ bool DataStreamZMQ::start(QStringList*)
       (dialog->ui->comboBox->currentText() + address + ":" + QString::number(port))
           .toStdString();
 
-  _zmq_socket.bind(_socket_address.c_str());
+  if (dialog->ui->radioConnect->isChecked())
+  {
+    _zmq_socket.connect(_socket_address.c_str());
+  }
+  else
+  {
+    _zmq_socket.bind(_socket_address.c_str());
+  }
 
   parseTopicFilters(topics);
   subscribeTopics();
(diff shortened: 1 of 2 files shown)


zdavkeos:
@jmount-boss I don't like the idea of forcing existing users to change their code. If anything, there should be an option to use `bind()` instead - and leave the default to `connect()`.  This pr would break our most common use-case of one publisher and multiple subscribers.

jmount-boss:
Agreed. Do you think adding a radio button within the subscriber ZMQ dialog with `connect()` as default would be acceptable? Or a different solution?

zdavkeos:
I think that a radio button and a blurb of documentation (maybe an image like the one on SO?) would add some nice functionality. That said, I think when most folks encounter non-typical flows they just code up a quick python script to to translate their situation into something that PJ understands.

jmount-boss:
No problem. I will get on that. Where would you suggest the blurb of documentation to go? And agree, a picture tells a thousand words.

Unfortunately, adding in a Python script/utility wasn't a satisfactory solution in this project. However, the flipping of bind and connect was. Hence, the PR.

zdavkeos:
Perhaps an image is overkill. Maybe the wording is enough - something like "Single Source / Multiple PJ's" and "Single PJ's / Multiple Sources"?

jmount-boss:
Added the radio buttons. Do you think option 1 or option 2 is the more informative blurb next to the radio buttons?

1. `Connect (Single Publisher / Multiple Subscribers)`; or
2. `Connect (Single Source / Multiple PlotJugglers`)

And similar for `Bind`

zdavkeos:
Just my opinion, but I think most users are going to be of the "PlotJuggler+DataProvider" mindset rather than "Publisher(s)+Subscriber(s)".

I think the work you've done so far looks good.

jmount-boss:
That is what I was thinking as well. I will alter it to be

- Connect (Single Data Source, Multiple PlotJugglers)
- Bind (Multiple Data Sources, Single PlotJuggler)

facontidavide:
Preferred a different solution. See this: f9808302f030eb79bc7cfb5bd43e73cdac47cb3f




Thanks for providing the idea

jmount-boss:
No problem. Glad to see it included

PR #890, opened 2023-12-03
Set the window title to the currently loaded log file name

stmoon:
When I use multiple PlotJuggler applications, I am always confused about which PlotJuggler window is associated with the log I am looking for.

To alleviate discomfort, I set the window title to the loaded log file name. It's a simple function, but I believe it's incredibly helpful.

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -1527,6 +1527,8 @@ std::unordered_set<std::string> MainWindow::loadDataFromFile(const FileLoadInfo&
 
       if (dataloader->readDataFromFile(&new_info, mapped_data))
       {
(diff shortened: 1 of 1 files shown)


zdavkeos:
@stmoon I pulled your branch and built it, and I do like your change. However, I think it needs to be taken a little further:

- There is already a feature to set the window title at startup (see mainwindow.cpp), this should not be overwritten.
- The title doesn't change when you start streaming; this would be a very nice addition.

Let me know what you think.

facontidavide:
This will not work when loading multiple files

Jofkos:
> This will not work when loading multiple files

Having the title be "{first filename} and others" would still be helpful.

facontidavide:
closing until the additional changes are introduces

PR #927, opened 2024-01-26
Fixed parsing JointState messages

peci1:
Without this fix, parsing of JointState messages is broken if `velocity` or `effort` fields are empty (which is allowed in the message definition).


facontidavide:
thanks a lot. Damn, not I need to do a new release :)

Issue #919, opened 2024-01-21
PlotJuggler does not build with fast-CDR version 2.x.x


manuelValch:
# Problem description

With latest version of fast-CDR: https://github.com/eProsima/Fast-CDR installed on the dev PC plotjuggler does not build.
Mainly because eProsima updated API name, namespaces and so on.

As a Sw developer I need to be able to compile plotJuggler with latest Fast-CDR so that I can have the data visualizer plugin from eProsima.

# Answer these questions (or I will ignore your issue)

1. Describe your platform / Operative System and the **version of PlotJuggler** you are using.
``` 
Static hostname: ubuntu-H610M-S2H-DDR4
       Icon name: computer-desktop
         Chassis: desktop
      Machine ID: f15bb09bbaa74f77be0ee3e9db35aca0
         Boot ID: ec841c97061f43a89768b3b36538327f
Operating System: Ubuntu 22.04.3 LTS              
          Kernel: Linux 6.5.0-14-generic
    Architecture: x86-64
 Hardware Vendor: Gigabyte Technology Co., Ltd.
  Hardware Model: H610M S2H DDR4
```

Version of PlotJuggler is the latest from main (3.8.8), commit hash: 92f7389b0112780cd9282522c1a348d5e54dd5c5
2. Compile and try the latest version in the **main** branch and check if the problem persists.

Yes it does persist.
```
/home/zorglub/plotjuggler_ws/src/PlotJuggler/plotjuggler_plugins/ParserROS/rosx_introspection/src/deserializer.cpp: In member function ‘virtual RosMsgParser::Span<const unsigned char> RosMsgParser::FastCDR_Deserializer::deserializeByteSequence()’:
/home/zorglub/plotjuggler_ws/src/PlotJuggler/plotjuggler_plugins/ParserROS/rosx_introspection/src/deserializer.cpp:196:21: error: ‘using element_type = class eprosima::fastcdr::Cdr’ {aka ‘class eprosima::fastcdr::Cdr’} has no member named ‘getCurrentPosition’; did you mean ‘get_current_position’?
  196 |   auto* ptr = _cdr->getCurrentPosition();
      |                     ^~~~~~~~~~~~~~~~~~
      |                     get_current_position
[...]

Issue #861, opened 2023-08-25
Plotjuggler compilation issue with FAST-CDR 1.1.0


MMarcus95:
Thanks for contributing to PlotJuggler. You are great!

# Problem description
Hello, I'm trying to compile Plotjuggler from source but it is not able to find fastcdr package. The installed version of FAST-CDR is 1.1.0. When compiling Plotjuggler (using either the last commit on main or the latest version 3.7.1) I'm getting the following error

![plotjuggler compilation issue](https://github.com/facontidavide/PlotJuggler/assets/52822119/5fa9ba29-e243-4bcc-9fd2-adc7bb6eb60e)

I'm actually having this issue since FAST-CDR 1.0.27. Since FAST-CDR 1.1.0 is the suggested version for FAST-DDS 2.11.0, it could be really nice to have Plotjuggler compiling with this version.

# Steps to reproduce (important)

- Platform: Ubuntu 22.04
- Install FAST-CDR 1.1.0
  - git clone https://github.com/eProsima/Fast-CDR.git -b 1.1.0
  - mkdir Fast-CDR/build && cd Fast-CDR/build
  - cmake .. && cmake --build . --target install
- Install Plotjuggler as described [here](https://github.com/facontidavide/PlotJuggler/blob/main/COMPILE.md#compile-in-linux)

paulBismuthListen:
Hello, 
Hhad the same on oracle 8 distribution.
Look like it is gone with cmake udpate, version I used is cmake-3.26.5

cheers.

facontidavide:
minimum cmake version I MUST support is 3.16.3, to be compatible with Ubuntu 20.04

PR #932, opened 2024-02-06
PlotJuggler with Fast-CDR-2.x.x

manuelValch:
Resolve: https://github.com/facontidavide/PlotJuggler/issues/919 https://github.com/facontidavide/PlotJuggler/issues/861

Reworked previous MR:
- removed cmake update to maintain compatibility with all version
- udpated cmake so that build does not fail if Fast-Cdr is found on dev PC

Added message if FastCdr is found on local PC:
```
-- [FastCdr] found, version: 2.1.3
```

If not found:
```
-- [FastCdr] not found, create shared libraries
```

# Tests

Tried to load some csv under `datasamples/rosbag2_test/`: 

image

Also able to stream data from fastDDS using eProsima plugin:

plotjugDDS

I can not test on old ros version but I hope it should be fixed now.


manuelValch:
ha no found out why it breaks ros2 as it has a pre-installed version of fastCdr

manuelValch:
After last commit, able to compile and run with fastCdr 1.1.1:
image

plotjuggler_plugins/ParserROS/rosx_introspection/CMakeLists.txt
```diff
@@ -9,15 +9,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
 
-find_package(fastcdr QUIET)
-
-if(NOT fastcdr_FOUND )
-    # Override Fast-CDR option: compile as static lib
-    SET(BUILD_SHARED_LIBS OFF CACHE BOOL "Create shared libraries by default")
-    add_subdirectory(3rdparty/Fast-CDR)
-    include_directories(3rdparty/Fast-CDR/include)
-endif()
-
```
facontidavide, line 20: I don't understand. If we are not compiling the library in 3rd party, how this even work?

facontidavide (changes requested):

plotjuggler_plugins/ParserROS/rosx_introspection/CMakeLists.txt
manuelValch: I moved it to the upper makefile:
plotjuggler_plugins/ParserROS/CMakeLists.txt

facontidavide: Yes, I believe you should move that code back to the other Cmake, thanks!!!

manuelValch: Moved back 3rdparty build where it was.
had to leave find(fastcdr QUIET) though otherwise build would fail with fastcdr installed on build machine i.e
this issue: https://github.com/facontidavide/PlotJuggler/issues/861

Also rebased on latest main which is why (I think) windows CI does not pass.

Issue #939, opened 2024-02-24
ULog float information printing broken


declancm:
# Problem description

ULog information messages of type float don't print correctly in the Info tab. They all display the same huge value. I have to use Pyulog to view the float information messages

![image](https://github.com/facontidavide/PlotJuggler/assets/90937622/8ae5f79b-eafe-4432-910a-c6b8a1ab1bb0)

# Answer these questions (or I will ignore your issue)

1. Describe your platform / Operative System and the **version of PlotJuggler** you are using.
    Windows 11
3. Compile and try the latest version in the **main** branch and check if the problem persists.
    Tested on 3.9.0 and read the release notes after - haven't been able to build master successfully.
5. Check if the problem can be reproduced using the dummy data created by the command line argument "-t" or one of the files in the folder "datasamples".
    There are no ULog files in the datasamples.
7. If it can't be reproduced with the dummy data, __please__ share the CSV file or the rosbag that can be used to reproduce the problem.
    Can't attach ULog files.

facontidavide:
Please provide a sample file to reproduce the issue

declancm:
https://drive.google.com/file/d/1xVuKQ5Ncl0fE78Wm5t_-E2K050dptbgk/view?usp=drive_link

declancm:
I'll work on a fix

declancm:
I've found the issue - the info values for most info message types are getting cast from the key strings instead of the value. I'll submit a pull request ASAP. This is a screenshot from the working info message:

![image](https://github.com/facontidavide/PlotJuggler/assets/90937622/1acaa0dc-2dcb-4d5f-ae30-369feb85b231)

facontidavide:
oh, thanks for finding this!

PR #946, opened 2024-03-06
Bugfix: Dereferencing wrong pointer for ULog information message value

declancm:
The Info tab is displaying the wrong values for non string types. The first part of the key (the type name) is getting de-referenced as the value for the information messages. I have created a string containing the value section of the message to use for de-referencing the raw data.

The Info tab previously:
image

The Info tab with this fix:
image

Note, the ULog used for testing only has float and char[] information messages. Those are not int information messages in the images.


declancm:
This pull request resolves issue #939

facontidavide:
thanks a lot!

PR #949, opened 2024-03-08
Updates to ZeroMQ plugin topic support

SebastianoTaddei:
Good morning,

first of all, thank you for this wonderful app. I started using it especially for the ability to stream data via ZeroMQ and while doing so I found some aspects of the plugin which (IMO) could be improved.

As this is my first pull request ever, please let me know if I needed to present it in different way or anything else you might suggest.

## Topics

Topic support was present, but did not ground incoming messages in said topics, resulting in overriding duplicate keys (e.g., the timestamp). In addition, no multipart messages were used, making it difficult to actually use it with a real PUB/SUB structure. 

I propose the following update to the plugin.

- The plugin now expects at least a two-part message, the first part is the topic (as string) and the second is the payload (the data you said PlotJuggler to expect).
- Using the topic as a key to a map of parsers, it maps correctly the message to its topic. Dynamic addition of topics as they come (according to the provided filter) is supported.
- The timestamp can be provided as part of the payload (as already present) or via a third part of the message. I believe this is useful if the data you are sending may not strictly contain the timestamp in the payload.

## Bug fixes

I found that the application would crash after a stop of the ZMQ plugin if it was started in bind mode, this was due a bug in the shutdown method.

### DIFF
--- a/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
+++ b/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
@@ -221,6 +221,7 @@ void DataStreamZMQ::receiveLoop()
     zmq::message_t recv_msg;
     zmq::recv_result_t result = _zmq_socket.recv(recv_msg);
 
+    // First part is the topic
     std::string topic;
     if (recv_msg.size() > 0)
     {
@@ -229,6 +230,7 @@ void DataStreamZMQ::receiveLoop()
       );
     }
 
+    // Second part is the payload
     if (recv_msg.more())
     {
       recv_msg.rebuild();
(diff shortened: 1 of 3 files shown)


facontidavide:
Hi,

can you please rebase and use pre-commit to check the formatting?

Also, I will only accept changes that are back compatible. Is this the case?

SebastianoTaddei:
Good evening,

apologies for the delay I struggled to understand what you meant by rebasing (I did not notice your last push). I can now confirm the pre-commit checks out.

In addition, I made some changes to ensure that the code should now be backward compatible, I have updated the test streamer found in the utilities of the plugin to test the new functionalities, and I have made sure the timestamp is now receive in seconds as double as it was everywhere else in the plugin.

What should I do now to update this pull request? Should I open a new one?

Issue #897, opened 2023-12-31
Compilation problem on Macbook air M1


ibeyzaekici:
Hello, I am encountering an issue while attempting to configure and build PlotJuggler. The error is related to the C compiler, and it seems to be failing during a simple test program compilation. The build process is encountering a "file cannot be mmap()ed" error with a specific path related to the Qt library.
1. *Platform / Operating System:** macOS Sonoma 14.2.1
2. **PlotJuggler Version: 3.8.4
3. **CMake Version: 3.28.1
4. **Xcode Version: Xcode 15.1

This is the error i am getting: 
<img width="1440" alt="Ekran Resmi 2023-12-31 21 02 35" src="https://github.com/facontidavide/PlotJuggler/assets/155296871/0fae5456-6971-4145-89fd-9fd1a1f8ad10">
<img width="1440" alt="Ekran Resmi 2023-12-31 21 02 39" src="https://github.com/facontidavide/PlotJuggler/assets/155296871/a021444c-d211-4e0e-abab-8eef011b1b3d">

I would be appreciated if you help me on that. Thank you.

NothanUmber:
Same problem here, if somebody finds a solution, mentioning it here would be great!

ubaldot:
Maybe this helps?

https://stackoverflow.com/questions/53633705/cmake-the-c-compiler-is-not-able-to-compile-a-simple-test-program

ubaldot:
I had the same problem and I tried the solution that I linked tonight and it didn't work.
That is, I opened the  `./src/PlotJuggler/CMakelist.txt ` file and add I added the following: 

`set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")`

just before `PROJECT.` (it is at the very beginning). 
Then, try to build again. It failed at 75% of build in the same linking error:

`ld: file cannot be mmap()ed, errno=22 path=/opt/homebrew/opt/qt@5/lib in '/opt/homebrew/opt/qt@5/lib'`

ubaldot:
**SOLVED**
1. Add the following to your `CMakeList.txt `file:

```
# Include PkgConfig module
find_package(PkgConfig REQUIRED)

# Use pkg_check_modules to find ZeroMQ
pkg_check_modules(ZEROMQ REQUIRED IMPORTED_TARGET libzmq)

# Include ZeroMQ directories
include_directories(${ZEROMQ_INCLUDE_DIRS})

```
[...]

PR #950, opened 2024-03-08
Added ZeroMQ and updated docs

ubaldot:
Necessary work to build from source code on macos 14.2.1 (and most likely newer and older). 
This PR also solves https://github.com/facontidavide/PlotJuggler/issues/897

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,15 @@ if(VCPKG_TOOLCHAIN)
     message(STATUS "BUILDING_WITH_VCPKG")
 endif()
 
+# Include PkgConfig module
+find_package(PkgConfig REQUIRED)
+
+# Use pkg_check_modules to find ZeroMQ
+pkg_check_modules(ZEROMQ REQUIRED IMPORTED_TARGET libzmq)
+
+# Include ZeroMQ directories
+include_directories(${ZEROMQ_INCLUDE_DIRS})
+
 find_package(ament_cmake QUIET)
 find_package(catkin QUIET)
 
--- a/COMPILE.md
+++ b/COMPILE.md
@@ -87,7 +87,7 @@ docker buildx build -o . .
 On Mac, the dependencies can be installed using [brew](https://brew.sh/) with the following command:
 
 ```shell
-brew install cmake qt@5 protobuf mosquitto zeromq zstd
+brew install cmake qt@5 protobuf mosquitto zeromq zstd zmqpp
 ```
 
 If a newer version of qt is installed, you may need to temporarily link to qt5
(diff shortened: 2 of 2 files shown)


facontidavide:
this PR caused a regression in Windows (see CI).

Please provide a solution or I will revert it

ubaldot:
I googled the error and it it seems that there is PkgConfig for windows. 
A possible solution could be to install the following package: 

`choco install pkgconfiglite`

But I have no Windows machine, so I cannot test if it actually solve. 
Would that be possible to run this solution over some GitHub actions Windows runner and see if it solve the issue?

facontidavide:
Yes, you should be able to run the CI in your own fork.

Nut please enable this in CMakeLists only when running in Macos, if you can.

ubaldot:
I managed to run the build locally, I fixed the build for Windows but that opened another problem with macos issue. This change is to fix the build for arm64 architecture, but the runner running the github workflow seems to use the x86 architecture only. 

However, I found this [recent blog article](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/) but it is not super clear to me how it should work. 

I'll try to investigate more...

ubaldot:
UPDATE: Since I managed to build PlotJuggler locally, it seems that something has changed either in PlotJuggler source code or in protobuf.

The error that I get around 93% of the build is the following:

```
/Users/runner/work/PlotJuggler/PlotJuggler/plotjuggler_plugins/ParserProtobuf/error_collectors.h:13:45: error: only virtual member functions can be marked 'override'
                const std::string& message) override;
                                            ^~~~~~~~
/Users/runner/work/PlotJuggler/PlotJuggler/plotjuggler_plugins/ParserProtobuf/error_collectors.h:16:47: error: only virtual member functions can be marked 'override'
                  const std::string& message) override;
                                              ^~~~~~~~
/Users/runner/work/PlotJuggler/PlotJuggler/plotjuggler_plugins/ParserProtobuf/error_collectors.h:31:45: error: only virtual member functions can be marked 'override'
                const std::string& message) override;
                                            ^~~~~~~~
/Users/runner/work/PlotJuggler/PlotJuggler/plotjuggler_plugins/ParserProtobuf/error_collectors.h:34:47: error: only virtual member functions can be marked 'override'
                  const std::string& message) override;
```

which seems that there is a `virtual` less or an `override` more and it seems an easy fix in the source code. However,  take that with a grain of salt since I am not a C++ guy. 

In-fact, I don't know if the issue is on protobuf or on PlotJuggler side, either way I don't dare to change the code by adding `virtual` or by removing `override`. By googling a bit, it seems there may be some relation with abseil but, again, I cannot say.

PR #951, opened 2024-03-08
ZMQ plugin improvements

SebastianoTaddei:
Good evening,

I have done what you suggested on the last pull request, it was closed so I guess I have to open a new one.

The pre-commit passed and the changes should be backward compatible from my tests. I will paste what the same message I had on the previous pull request.

## Topics

Topic support was present, but did not ground incoming messages in said topics, resulting in overriding duplicate keys (e.g., the timestamp). In addition, no multipart messages were used, making it difficult to actually use it with a real PUB/SUB structure.

I propose the following update to the plugin.

The plugin now can receive a two-part message, the first part is the topic (as string) and the second is the payload (the data you said PlotJuggler to expect).
Using the topic as a key to a map of parsers, it maps correctly the message to its topic. Dynamic addition of topics as they come (according to the provided filter) is supported.
The timestamp can be provided as part of the payload (as already present) or via a third part of the message. I believe this is useful if the data you are sending may not strictly contain the timestamp in the payload.

## Bug fixes

I found that the application would crash after a stop of the ZMQ plugin if it was started in bind mode, this was due a bug in the shutdown method.

### DIFF
--- a/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
+++ b/plotjuggler_plugins/DataStreamZMQ/datastream_zmq.cpp
@@ -1,13 +1,14 @@
 #include "datastream_zmq.h"
 #include "ui_datastream_zmq.h"
 
-#include <QMessageBox>
+#include "PlotJuggler/messageparser_base.h"
 #include <QDebug>
-#include <QSettings>
 #include <QDialog>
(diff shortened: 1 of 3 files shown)


lu-maca:
Hi @facontidavide, are you planning to merge this PR?

Thanks! :)
Luca

PR #954, opened 2024-03-16
CSV: Allowing to parse date as extended ISO 8601

Jofkos:
When working with data from different sources, one may encounter a various number of different date formats.

Examples include:
- yyyy-MM-dd hh:mm:ss
- yyyy-MM-dd hh:mm:ss.zzz
- yyyy-MM-dd hh:mm:ss.zzzzzz (not even supported with format strings)
- yyyy/MM/dd hh:mm:ss
- yyyy/MM/dd hh:mm:ss.zzz
- yyyy/MM/dd hh:mm:ss.zzzzzz (not even supported with format strings)

These are all supported by Qt's `Qt::ISODateWithMs`, which thus allows for instantly loading different csv's without having to adjust the date format every time. This pull request suggests using ISO8601 parsing when no format string has been entered.


facontidavide:
with the latest change that is smarter in terms of auto-detecting the time format, this change is probably not needed anymore.

Thanks

PR #956, opened 2024-03-23
Udp-socket-thread

Gaza-Israel:
Switched from QSocket to default linux sockets to implement the receive of packets in a separeted thread.

This solves an issue with high frequency data where plotjuggler would lose several packets while updating the plots. This issue only happens while livestreaming whithout pause, if you pause the window it receives normally all the data, but as soon as it is unpaused the issue returns.

I'm aware that this solution will lead to problems with windows implementation, but for now I dont have the time to implement with other OS compatibility.

Tested on Ubuntu 22.04..4 LTS

### DIFF
--- a/plotjuggler_plugins/DataStreamUDP/udp_server.cpp
+++ b/plotjuggler_plugins/DataStreamUDP/udp_server.cpp
@@ -23,16 +23,9 @@ THE SOFTWARE.
 #include <QSettings>
 #include <QDialog>
 #include <mutex>
-#include <QWebSocket>
 #include <QIntValidator>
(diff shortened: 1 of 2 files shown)


unicoper:
pretty cool work, works stable for me, congrats
now I can see my signals at 65mhz without interruptions

facontidavide:
as you said, I can't merge this unless Window is supported too.

facontidavide:
I love the idea of using a separate threadm but can we avoid using raw socket and use Qt instead?

facontidavide:
closing for conflict. also, any solution proposed should work in Windows too

PR #959, opened 2024-04-01
CSV: added option to skip invalid rows

Jofkos:
CSV files may contain invalid rows (i.e., rows that have the wrong number of columns). In some cases, those are not expected and thus not loading the entire file (and displaying a warning) may make sense. This is what is currently done. In other cases, however, due to how the CSV files are generated, invalid rows may be expected and still being able to load such a file without having to first manually remove them would thus be helpful.

This pull request does exactly that: It adds a checkbox to just ignore rows which don't have as many values as there are columns. It is unchecked by default, preserving the current behaviour.

Screenshot 2024-04-01 at 16 26 57

### DIFF
--- a/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
+++ b/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
@@ -301,6 +301,8 @@ int DataLoadCSV::launchDialog(QFile& file, std::vector<std::string>* column_name
       settings.value("DataLoadCSV.useDateFormat", false).toBool());
   _ui->lineEditDateFormat->setText(
       settings.value("DataLoadCSV.dateFormat", "yyyy-MM-dd hh:mm:ss").toString());
+  _ui->checkBoxSkipInvalidRows->setChecked(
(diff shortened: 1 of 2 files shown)


facontidavide:
I wonder who wants to load a corrupted CSV...

Jofkos:
I made that change because it is useful.

Here's one use-case: A CSV gets written to by an application which is connected to a physical instrument. You don't want to stop the experiment, but want to quickly inspect the data as it is running. When you grab the CSV file, the last line is incomplete. Sure, this is not a problem: just delete the last line. When you have to do this multiple times a day, having the ability to just drag in the files rather than having to open them in a separate application first is a big win.

Here's another one: Some process may just generate CSV where some lines are incomplete. This could be because of fickle communication, interference or a non-prioritized telemetry printing being interrupted improperly. Just because that's the case, does that mean the telemetry that does get through is worthless?

pkess:
I have a similar use case as @Jofkos . But i did not see this pr before. So i already created a similar pr: #1024 .

facontidavide:
not going to accept this for philosophical reasons

PR #960, opened 2024-04-04
Added support for empty messages

jbendes:
This commit adds support for zcm messages that are empty. Before the messages would just be dropped and not displayed in PlotJuggler. Now they will show up at the correct timestamp with no contents and a numeric value of 0 for each message that exists


facontidavide:
please apply the correct formatting using **pre-commit**

facontidavide:
also, are you sure that this is a good idea? how do we distinguish between "zero as null" and actual zeros?

jbendes:
How do I do that?

jbendes:
That actually can't be a thing. A message in zcm is like a c-struct. There are always fields inside the message that have values. The message itself can't have a value. Does that make sense?

jbendes:
@facontidavide done!

Issue #968, opened 2024-04-23
Misalignment in Time Series Data Due to Differing Trajectory Sizes with reactive scripts


salmagro:
# Issue: Misalignment in Time Series Data Due to Differing Trajectory Sizes with reactive scripts

## Problem Description

Currently, PJ uses the nearest time index for data retrieval (`TimeseriesRef::atTime(double t)`), which can lead to inaccuracies when analyzing time series data of different sizes. This method may inadvertently use outdated data towards the end of a data set, affecting analysis reliability.

## Example of the Problem
Please found some dummy data here: [DummyData_2024-03-19.csv](https://github.com/facontidavide/PlotJuggler/files/15073491/DummyData_2024-03-19.csv)

Expected           |  Currently
:-------------------------:|:-------------------------:
![2024-04-23_09-38_trajectorycropped](https://github.com/facontidavide/PlotJuggler/assets/52298571/bdca386d-5096-4897-a2a8-3c14c7736308) |![2024-04-23_09-36_trajectory_error](https://github.com/facontidavide/PlotJuggler/assets/52298571/d73c0a33-20fe-49ff-a9cb-9c59311dcf1a)

## Implemented Workaround

I have exposed the internal index through `TimeseriesRef::getIndexAtTime(double t)` to ensure that indices match when correlating data points across different series. This ensures that comparisons and analyses use data points from the same measurement intervals.
```diff
diff --git a/plotjuggler_base/include/PlotJuggler/reactive_function.h b/plotjuggler_base/include/PlotJuggler/reactive_function.h
index 601cfe43..0e903408 100644
--- a/plotjuggler_base/include/PlotJuggler/reactive_function.h
+++ b/plotjuggler_base/include/PlotJuggler/reactive_function.h
@@ -27,6 +27,8 @@ struct TimeseriesRef
 
   double atTime(double t) const;
 
+  int getIndexAtTime(double t) const;
+
   unsigned size() const;
 
   void clear() const;
diff --git a/plotjuggler_base/src/reactive_function.cpp b/plotjuggler_base/src/reactive_function.cpp
index d9c713a3..b15a73f0 100644
--- a/plotjuggler_base/src/reactive_function.cpp
+++ b/plotjuggler_base/src/reactive_function.cpp
@@ -114,6 +114,7 @@ void ReactiveLuaFunction::prepareLua()
   _timeseries_ref["at"] = &TimeseriesRef::at;
[...]

PR #969, opened 2024-04-23
Enhance Data Retrieval Accuracy with MatchType in TimeseriesRef

salmagro:
This MR introduces the `MatchType` enumeration to `TimeseriesRef`, providing the option to choose between exact and nearest time matching. Changes include updates to the `atTime` and `getRawIndexAtTime` methods to support this new functionality.

For a detailed problem description see: https://github.com/facontidavide/PlotJuggler/issues/968

### DIFF
--- a/plotjuggler_base/include/PlotJuggler/reactive_function.h
+++ b/plotjuggler_base/include/PlotJuggler/reactive_function.h
@@ -27,6 +27,8 @@ struct TimeseriesRef
 
   double atTime(double t) const;
 
+  int getRawIndexAtTime(double t) const;
+
   unsigned size() const;
 
   void clear() const;
--- a/plotjuggler_base/include/PlotJuggler/reactive_function.h
+++ b/plotjuggler_base/include/PlotJuggler/reactive_function.h
@@ -25,7 +25,7 @@ struct TimeseriesRef
 
   void set(unsigned index, double x, double y);
 
-  double atTime(double t) const;
+  double atTime(double t, MatchType match_type) const;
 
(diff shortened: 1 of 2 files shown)


facontidavide:
look at CI...

salmagro:
I have resolved the compilation issues. However, I have a question regarding the proposal to pass an enum class `MatchType` to the `atTime` method. 
Do we need to make any modifications when exposing this to Lua in this line https://github.com/facontidavide/PlotJuggler/blob/main/plotjuggler_base/src/reactive_function.cpp#L116? 
How should the enum class MatchType be defined from Lua?

plotjuggler_base/include/PlotJuggler/reactive_function.h
```diff
@@ -25,7 +31,10 @@ struct TimeseriesRef
 
   void set(unsigned index, double x, double y);
 
-  double atTime(double t) const;
+  double atTime(double t, MatchType match_type) const;
```
firemark, line 34: Adding new argument breaks older scripts. The best way it should be a default value but ain't sure it's possible with lua.

firemark:
> How should the enum class MatchType be defined from Lua?

Maybe you should define standalone functions [like Uncle bob once said](https://x.com/unclebobmartin/status/1114137614377005057). In this way you should add three functions:

* `getRawIndexAtExactTime`
* `getRawIndexAtTime`
* `getAtExactTime`

facontidavide:
the timeseries class will be completely refactored in the next months and will include native types. closing this.

Thanks anyway :smile:

PR #970, opened 2024-04-24
Use regexp when applying visibility filter

gh-not-good:
This is good when we need to filter the list of signals based on regular expression rather than just letters in the names of the signals.

Please note this PR is start of the discussion if this functionality is even wanted. I am ready to make appropriate changes in the format if needed. Thanks!


gh-not-good:
I would like to kindly ask about the guidance on how to pass the *AppVeyor* check.

Issue #839, opened 2023-06-07
UDP JSON stream timestamp setting not saved with layout file


Bartimaeus-:
# Problem description

Configuration of the timestamp field for UDP streaming with JSON is not maintained when loading a layout file
![image](https://github.com/facontidavide/PlotJuggler/assets/2954254/df37684b-fdb2-44c1-b5b3-bcf7e6ab9669)

My use case for this is trying to launch PlotJuggler from another program that will stream data to it. 

# Steps to reproduce (important)

- Set up UDP steam with JSON and a timestamp field in PlotJuggler 
- Save layout file
- Close PlotJuggler and open a new instance and load layout file
- UDP stream dialog will not have timestamp field setting

Example layout and UDP streaming Python test file included below
[PJ_issue_UDP_JSON_layout_file.zip](https://github.com/facontidavide/PlotJuggler/files/11679777/PJ_issue_UDP_JSON_layout_file.zip)

PR #971, opened 2024-04-25
Save NlohmannParser (JSON) settings

MichelJansson:
With this change, the NlohmannParser (json/bson/cbor/msgpack parsers) options are saved to the user settings file, making especially data streaming easier where we no longer have to specify this option after each restart of the app.

This PR partially solves #839: Settings are now saved, but not as part of the layout file.

### Changes:
- Settings saved when parser is instantiated.
- Settings loaded when the options widget is accessed.

### Design reasoning:
I debated over extending the `ParserFactoryPlugin` interface with `void saveSettings()` and `void loadSettings()` to make this feature explicit and controllable by the caller, that also would improve the interface for future plugins.
In the end I opted for this simpler more contained solution not to change the architecture, but with the downside of settings being loaded multiple times and both load and save being a side-effect.

I would be happy update the PR with the interface route if you think that is a better approach.

Thanks

### DIFF
--- a/plotjuggler_app/nlohmann_parsers.h
+++ b/plotjuggler_app/nlohmann_parsers.h
@@ -10,6 +10,7 @@
 #include "nlohmann/json.hpp"
 #include "PlotJuggler/messageparser_base.h"
 #include <QDebug>
+#include <QSettings>
 
 using namespace PJ;
 
@@ -113,11 +114,14 @@ public:
   NlohmannParserCreator()
   {
     _checkbox_use_timestamp = new QCheckBoxClose("use field as timestamp if available");
+    loadSettings();
   }
 
   template <typename ParserT>
   MessageParserPtr createParserImpl(const std::string& topic_name, PlotDataMapRef& data)
   {
+    saveSettings();
+
     std::string timestamp_name = _checkbox_use_timestamp->lineedit->text().toStdString();
     return std::make_shared<ParserT>(
         topic_name, data, _checkbox_use_timestamp->isChecked(), timestamp_name);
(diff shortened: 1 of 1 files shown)


plotjuggler_app/nlohmann_parsers.h
```diff
@@ -113,21 +114,43 @@ class NlohmannParserCreator : public ParserFactoryPlugin
...
+    _checkbox_use_timestamp->lineedit->setText(field);
+  }
+
+  virtual void saveSettings()
+  {
+    QSettings settings;
+    settings.setValue("NlohmannParser.timestampEnabled", _checkbox_use_timestamp->isChecked());
+    settings.setValue("NlohmannParser.timestampFieldName", _checkbox_use_timestamp->lineedit->text());
+  }
+
```
facontidavide, line 153:
```suggestion
  virtual void loadSettings()
  {
    QSettings settings;
    QString prefix = QString("NlohmannParser.") + QString(encoding());
    bool checked = settings.value(prefix + ".timestampEnabled", false).toBool();
    QString field = settings.value(prefix + ".timestampFieldName", "").toString();

    _checkbox_use_timestamp->setChecked(checked);
    _checkbox_use_timestamp->lineedit->setText(field);
  }

  virtual void saveSettings()
  {
    QSettings settings;
    QString prefix = QString("NlohmannParser.") + QString(encoding());
    settings.setValue(prefix + ".timestampEnabled", _checkbox_use_timestamp->isChecked());
    settings.setValue(prefix + ".timestampFieldName", _checkbox_use_timestamp->lineedit->text());
  }

```

facontidavide: shall we have different parameters, based on encoding?

MichelJansson: Sure, makes sense! That was also a neat way of doing it without duplication.

MichelJansson: Hmm, that code change actually crashes on my computer. I'm assuming `encoding()` of the derived class is not callable from the base class constructor.

Perhaps `loadSettings()` could be removed from the constructor, which I mostly added for "good measure", in case the options were to be used within the factory class in the future.

MichelJansson: I solved that by moving the `encoding()` function up to `NlohmannParserCreator`. Hope that is OK!

PR #974, opened 2024-04-26
Draft: Enable TimeTracker when streaming

gh-not-good:
Needs testing in the streaming mode yet.


gh-not-good:
It's draft but ready for review.

facontidavide:
there is a reason why this must be OFF during streaming. We don't want to trigger the republish plugin

gh-not-good:
Is there any other way how to see the actual values of the signals in the plot during the streaming?

PR #979, opened 2024-06-04
statistics: add avg interval

dakejahl:
It's nice to know at a glance the rate of which a topic is logged


zdavkeos:
I think this is a good idea, the changes look good. One nit would be the name `Avg Interval` might not be clear enough. Is there a way to add a tooltip (e.g "Average time per sample")? Or change the other fields to be specific, e.g "Minimum Value", "Maximum Value", etc.

dakejahl:
I've been using this for a while and it's really helpful. Can we merge it?

PR #986, opened 2024-06-26
[ROS-O] drop broken test hook from Fast-CDR import

v4hn:
The tests were not imported but still referenced in the cmake depending on the system cmake version.

My build (Debian testing, ROS-O)  just got stuck on:

```
CMake Error at plotjuggler_plugins/ParserROS/rosx_introspection/3rdparty/Fast-CDR/CMakeLists.txt:136 (add_subdirectory):
  add_subdirectory given source "test" which is not an existing directory.
```

It looks like you imported the library without its tests, but did not remove the cmake reference.


v4hn:
I just noticed the same mistake exists in https://github.com/facontidavide/rosx_introspection , not sure about others.

facontidavide:
thanks

PR #996, opened 2024-08-08
Feature: Added option to merge identically named timeseries across files

Jofkos:
_Note: If any part of this pull-request is not up to the expected standards, preventing it from being merged, I would appreciate any feedback so that I can iterate on it until it is!_

## Background
Depending on the software used to collect data during experiments, the resulting logs can be fragmented. If that is the case, one can, currently, visualize them all at the same time in PlotJuggler by just giving them different prefixed.

This approach has one main drawback, however: Every timeseries is present as often as the number of log fragments one is dealing with. As a result, layouts don't work and custom series are cumbersome (need to be created for every <prefix>/timeseries).

## Changes
This pull request adds an option in the 'multifile prefix' dialog, that, when ticked, merges timeseries of the same name of all files loaded. In the use-case described before, this can be used in two ways: (1) By setting all prefixes to empty strings, the resulting timeseries in PlotJuggler behave as if coming from one individual file. (2) By grouping the timeseries into different prefixes, multiple fragmented datasets can be visualized simultaneously.

Additionally, this pull requests adds a 'clear already present data' checkbox, that is only active when 'merge identically named timeseries' is, too, and does as the name suggests. When not ticked, newly loaded timeseries are merged with the ones already present in PlotJuggler. When ticked, all points in existing timeseries are clearer before importing the new data.

## Examples and Screenshots
For example, this configuration:
image

where the files all contain overlapping timeseries would merge the content of those.

Unchecking the box results in the current behaviour, where the last file loaded is the one whose timeseries with clashing names survive.
image

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -1373,7 +1373,7 @@ bool MainWindow::loadDataFromFiles(QStringList filenames)
 {
   filenames.sort();
   std::map<QString, QString> filename_prefix;
-  bool mergeFiles = false;
+  bool mergeFiles = false, clearExisting = true;
 
   if (filenames.size() > 1 || ui->checkBoxAddPrefixAndMerge->isChecked())
   {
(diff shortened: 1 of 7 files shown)


Valle125:
A very handy feature!

facontidavide:
I think the new "merge" checkbox does this

PR #1003, opened 2024-09-07
Do not convert log extensions to lowercase

muramura:
The log files in ArduPilot have an extension of uppercase BIN. 
I want to make decisions based on the string set in the extension.

ArduPilot PLUGIN
https://github.com/ArduPilot/plotjuggler-apbin-plugins


facontidavide:
i don't think that using upper/lower case as discriminant is a good idea.

PR #1009, opened 2024-09-21
Fix MQTT plugin (#1006, #1008)

### LINKED ISSUE #1006: MQTT subscriber 3.9.0 windows does not show any topics at all.
# Problem description

Running Windows 10.

Just upgraded from 3.6.0 to 3.9.0

But now MQTT dialog does not list any of the (persistent) topics. Closed, run 3.6, showing all topics. (needed 3.9 to be able to select more than 1 topic from the list).

So back to 3.6.0 for now.

### LINKED ISSUE #1008: MQTT plugin freezes PlotJuggler on Linux.
The MQTT plugin causes PlotJuggler 3.9.1 to freeze on Linux (Ubuntu 22) if the connect button is hit repeatedly. If this happens PlotJuggler doesn't react to any input. The program must be terminated e.g. by using the task manager. 

Screenshot from 2024-09-21 20-24-19

Valle125:
This pull request fixes issues #1006 and #1008.

On Linux the MQTT plugin freezes PlotJugger see issue #1008. The program flow seems to be stuck within mosquitto_connect_bind_v5() on the second connection attempt. This is solved by starting with a fresh mosquitto instance at each connection attempt.

On Windows PlotJuggler doesn't show any topics after establishing a connection, see issue #1006. mqtt_client.cpp does not check the return code of mosquitto_loop_start(), which may return MOSQ_ERR_NOT_SUPPORTED on Windows. This causes the plugin to assume that the connection has been established, but in reality the mosquitto event loop is not processed. This causes the user to think that a connection was established but no topics are displayed. The plugin worked in the past with no recent changes to the source code, so what causes the problem?  It seems that libmosquitto has at some point dropped threaded support on Windows (see https://github.com/eclipse/mosquitto/issues/2707). Threaded support will be added again in the future (maybe 2.1?). To solve this problem mosquitto gets it's own thread if mosquitto_loop_start() fails. 

To make mqtt_client.cpp more robust all mosquitto_xxx return codes are checked. To increase the user experience some QMessageBox messages have been added.

fixes #1006 
fixes #1008 

Kind regards,
Valentin

### DIFF
--- a/plotjuggler_plugins/DataStreamMQTT/mqtt_client.cpp
+++ b/plotjuggler_plugins/DataStreamMQTT/mqtt_client.cpp
@@ -2,6 +2,7 @@
 #include <QDebug>
 #include <QMessageBox>
 #include <QString>
+#include <QtGlobal>
 
 #ifdef WIN32
 #include <windows.h>
@@ -52,11 +53,6 @@ void message_callback(struct mosquitto* mosq, void* context,
 MQTTClient::MQTTClient()
 {
   mosquitto_lib_init();
-  _mosq = mosquitto_new(nullptr, true, this);
-
-  mosquitto_connect_v5_callback_set(_mosq, connect_callback);
-  mosquitto_disconnect_callback_set(_mosq, disconnect_callback);
-  mosquitto_message_v5_callback_set(_mosq, message_callback);
 }
 
 MQTTClient::~MQTTClient()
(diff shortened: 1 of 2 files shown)


Valle125:
Hi, 
would somebody mind to review my changes? Plotjuggler is a great program and I'd love to see the MQTT plugin to work again. Quiet some people seem to have the issue (#1006, https://github.com/PlotJuggler/plotjuggler-mqtt/issues/4).
Best,
Valentin

PR #1022, opened 2024-11-15
Zcm loader prefix

jbendes:
Supporting a prefix on data loaded and streamed into plotjuggler via zcm


jbendes:
NEVERMIND. Just saw that this is built-in now haha

PR #1024, opened 2024-11-18
Allow user to skip malformed lines in CSV

pkess:
Replace error message and abort action in csv import plugin with a question for the user how to proceed


pkess:
Looks like i based my work on an outdated fork. I will update it again.

pkess:
> Looks like i based my work on an outdated fork. I will update it again.

I rebased and updated the changes now.

facontidavide:
no, of your CSV is malformed, it is your problem

pkess:
So i created this pr more than 6 month ago without any feedback and now you are closing this with only this small feedback? I am not happy about this.

The reason for this is very common  IMO: When data is converted from any file format to the CSV format the changes are written line by line. While this can take a long time you might want to get a preview to the data and try to load all data that was already converted. When i did this in the past the csv importer always complained about one line not being complete (the last line in the file that could just be skipped). And the csv failes to load.
So what i am doing is i copy the file that is being converted into, open it in a text editor and remove the last line. Then i can load it with plotjuggler. With this pr i could just load the file and there will be a message that i can ignore.

So @facontidavide i am asking you to think about this again.

PR #1026, opened 2024-11-22
Reload layout

jbendes:


facontidavide:
remerge and fix CI, and I will merge it

facontidavide:
so, you do two changes: reload layout and change in ZMQ.

And of course, why should you add a description of what you do, a screenshot of the change in the GUI, etc?

Please be a little less lazy if you want your PR to be merged.

jbendes:
Hey @facontidavide, it sounds like you have a desired structure for a pullreq. Is that documented anywhere in the repo? Typically an open source repo would have a contributing.md or something similar with an outline of whatever you care about having in the merge request. I personally don't think about merge requests the way you do. I think of them as the beginning of a conversation around whether or not the primary maintainer wants this feature in their repo.

facontidavide:
>  Is that documented anywhere in the repo? 

I should certainly provide more instructions / guidelines for contributors!!!

Anyway, as a rule of thumb, the bare minimum that anyone will agree with is:

- on PR should apply only one specific change
- PR with no descriptions should be either trivial (not this case) or... no really, you must add a description.

Additionally, if a PR changes the appearency of the GUI, a screenshot will be very welcome.
I am not going to spent time compiling your branch just to check it out.

PR #1032, opened 2025-01-03
Build improvements

stefanct:
The first commit fixes a problem that I have reported upstream as well (https://github.com/eProsima/Fast-CDR/pull/259). With this change it is possible to cross-compile PJ with mingw for windows.

The second patch allows for building PJ statically. It's meant as a base for discussion. The resulting binary is naturally huge but easy to distribute.


facontidavide:
check #1041 and see if this is still necessary

PR #1037, opened 2025-01-23
DataLoadMCAP: add support to mcap logTime

lucabart97:
The DataLoadMCAP plugin support only the publishTime of the mcap message. Add a radio button in the load dialog to select the logTime or publishTime.


facontidavide (changes requested):
I believe that a Radio button is better than a checkbox, to make clearer to the user what the alternatives are.

lucabart97:
Got it!
Selection_008

PR #1043, opened 2025-02-02
Add parsing support for tsl messages

simonsag96:
We at TUM Autonomous Motorsport developed our own time series logging library - comparable to Data Tamer and pal_statistics. By the way, Kudos to DataTamer, it's an awesome library. 
However, we already started developing our library before DataTamer was around. 
And even then,  it did unfortunately not fit all the requirements we had (e.g. access from python....).
 
Our library called [TSL](https://github.com/TUMFTM/tsl) is now also OpenSource.
Therefore, I would really like it if the message definitions were supported officially by PlotJuggler and not just by our fork.

Let me know what you think of the PR. For the implementation of the message parsing, I have oriented myself on your implementation of the pal_statistics and data tamer parser. The implementation has been tested for months internally without any problems.


simonsag96:
Regarding the comments:
I intentionally did not parse the timestamp, but I could add an extra timeseries if you prefer.

facontidavide:
merging it, but I take no responsibilities :smile:

simonsag96:
Awesome, thanks!

Just link me on the issue if something comes up in the future. I am willing to maintain it as well in case!

PR #1045, opened 2025-02-15
Pan plots with arrow keys

zdavkeos:
The "scale" of each move is set arbitrarily for now. I wanted to see what folks thought.


facontidavide:
no, thanks

PR #858, opened 2023-08-17
Add multicast support to DataStreamUDP.

agoessling:
This adds the ability to specify a address other than 127.0.0.1 when starting DataStreamUDP.  If the address is a multicast address the UDP socket is configured to join the multicast group.

PR #1050, opened 2025-03-03
Fix IPv6 multicast

bryceschober:
Trying IPv6 multicast, I could get nothing to work. After much trial & error, I got this solution working, including using the `udp_client.py` tool as a test data source with the documented multicast groups for both IPv4 and IPv6. I noticed the previous #858, @agoessling would you like to re-test your use-case?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ _deps
 # Clangd
 .cache
 /snap/*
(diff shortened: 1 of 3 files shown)


facontidavide:
looks good to me. thanks!

PR #1107, opened 2025-07-03
Feature/cvs skip corrupted lines

Valle125:
# Summary
This pull request enhances the behavior of the DataLoadCSV::readDataFromFile function and consists of two commits. 
 * The first fixes the line numbers displayed within the QMessageBox windows. The line numbers where incorrect when the CVS file contained empty lines. 
 * The second commit changes the behavior of the procedure when it encounters incomplete/corrupted lines. Incomplete lines are now skipped and reported to the user at the end. The user can than choose to abort or continue.

# Detail
Below you can find some test files that demonstrate the incorrect behavior of the old code which is fixed by the first commit. The content of the res_failX.txt and res_okX.txt files correspond to the detail text output of the QMessageBox windows of the old/new code for the corresponding testX.csv file. Since there is a single empty line within all files the old version of the code displays a line number that is off by one.

Time not monotonic (column a, row 4): [test1.csv](https://github.com/user-attachments/files/21041740/test1.csv)
Column incomplete (row 4): [test2.csv](https://github.com/user-attachments/files/21041743/test2.csv)
Time parsing error (row 4): [test3.csv](https://github.com/user-attachments/files/21041745/test3.csv)
[res_fail1.txt](https://github.com/user-attachments/files/21041741/res_fail1.txt)
[res_fail2.txt](https://github.com/user-attachments/files/21041744/res_fail2.txt)
[res_fail3.txt](https://github.com/user-attachments/files/21041746/res_fail3.txt)
[res_ok1.txt](https://github.com/user-attachments/files/21041747/res_ok1.txt)
[res_ok2.txt](https://github.com/user-attachments/files/21041748/res_ok2.txt)
[res_ok3.txt](https://github.com/user-attachments/files/21041794/res_ok3.txt)

The changes in behavior introduced by the second commit are shown in the picture and files below. The picture can be reproduced by loading the test4.csv file. It contains multiple empty and incomplete lines. Once all data has been parsed, this window appears and one can choose to abort or continue.
image
Test multiple empty lines and incomplete rows: [test4.csv](https://github.com/user-attachments/files/21041749/test4.csv)
[res_ok4.txt](https://github.com/user-attachments/files/21041750/res_ok4.txt)

### DIFF
--- a/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
+++ b/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
@@ -610,6 +610,7 @@ bool DataLoadCSV::readDataFromFile(FileLoadInfo* info, PlotDataMapRef& plot_data
 
   int linenumber = 1;
   int samplecount = 0;
+  QVector<std::tuple<long, long, QString>> skip_list;
   while (!in.atEnd())
   {
     QString line = in.readLine();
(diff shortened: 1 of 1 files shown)


Jofkos:
This is indeed an important yet missing aspect of the CSV loader, but previous PRs (#1024 or #959, potentially others) suggesting solutions to this have been closed ["for philosophical reasons"](https://github.com/facontidavide/PlotJuggler/pull/959#issuecomment-2970588379). Just FYI.

If this was something that could potentially be merged, I feel like for this to be maximally useful, an option to skip this dialogue and accept the risk to always skip corrupted CSV lines would be very useful to have (if you already know the corruption and in what way it will be corrupted, the dialogue is just additional friction).

Valle125:
Thank you for your comment @Jofkos . This is also a feature that I and my colleagues are looking forward to. I'm currently working with an embedded device that is regularly power cycled and that generates log files that are hundreds of megabytes large. The device appends data to existing log files after a power cycle until the log file reaches a maximum size. This means that occasionally when the device is power cycled during a write operation the last line may be incomplete. Hence the file may contain quit a few incomplete lines with otherwise flawless data. Removing those lines by hand is not practical for large files.

I'm hopeful that we can find a solution that suits everybody.

facontidavide:
merged with some changes

PR #1111, opened 2025-07-08
Can now show only the current time range in XY graphs

### LINKED ISSUE #1104: Display only the data for the current time interval in the X/Y plots.
It would be cool to be able to show only the data for the currently selected time interval in the X/Y plots. Maybe as an option. We could add a check box in the "Edit Curves" dialog to enable this feature.

I'm interested to do a pull request with such a change, I've created this ticket just to see if it makes sense to the people in charge...

pvalsecc:
Closes #1104


facontidavide:
thanks a lot, I will try this soon. I may probably apply some changes

PR #1117, opened 2025-07-10
Added option to skip splash screen to settings ini file

thenoname:


facontidavide:
Have you tried it? I don't think this will work

facontidavide:
there is already a way to prevent that using -t argument at launch. it is enough.

PR #1122, opened 2025-07-15
Add no splash to settings

thenoname:
Adds an optional way to disable the splash screen on load. Useful for instances where the launch arguments cannot be easily modified.


facontidavide:
I don't understand. How is that setting updated to a certain value?

thenoname:
Apologies for not better explaining in the description. 

My thought was that a user could just add `no_splash=true` to their plotjuggler-3.ini file manually if they wanted to use this. Perhaps that is not a good method?

facontidavide:
I don't think it is, you should probably add an option in preferences
dialog

PR #1129, opened 2025-07-24
Solve the PALStatistics message parsing in case several topic are published

MaximilienNaveau:
## Introduction

This PR introduce a bugfix in the parsing of the PAL-Statistics topics.

## Description

In case several topic are published like for example:
- `/estimator/introspection_data/full`
- `/estimator/introspection_data/names`
- `/estimator/introspection_data/values`
- `/controller/introspection_data/full`
- `/controller/introspection_data/names`
- `/controller/introspection_data/values`
Then the behavior of the ROS parser is wrong as the static variable does not distinguish the different topics individually leading to a wrong parsing of the names versus data.

Therefore I created a map containing the prefix: `/controller/introspection_data` or `/estimator/introspection_data` where I store the topic names and versions.
Hence this map is filled up in the `parsePalStatisticsNames` and reused later depending on which topic the current message is read from.
Tried on Ubuntu22.04 humble.

### DIFF
--- a/plotjuggler_plugins/ParserROS/ros_parser.cpp
+++ b/plotjuggler_plugins/ParserROS/ros_parser.cpp
@@ -552,10 +552,12 @@ void ParserROS::parseDataTamerSnapshot(const std::string& prefix, double& timest
   DataTamerParser::ParseSnapshot(dt_schema, snapshot, callback);
 }
 
-static std::unordered_map<uint32_t, std::vector<std::string>> _pal_statistics_names;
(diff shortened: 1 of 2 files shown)


facontidavide:
merge. please remeber formatting the next time

MaximilienNaveau:
It seems you closed this PR rather than merge it. So I wonder if these changes were applied?
I don't see any commit in that direction.

MaximilienNaveau:
I pushed a format commit

PR #1132, opened 2025-07-27
Bugfix/mnaveau/pal statistics several topics

MaximilienNaveau:
## Introduction

This PR introduce a bugfix in the parsing of the PAL-Statistics topics.
It's the same PR as this closed one: https://github.com/facontidavide/PlotJuggler/pull/1129
@facontidavide I recreate it here as I cannot reopen the other one.
Hope it's not too much of a trouble.

## Description

In case several topic are published like for example:
- `/estimator/introspection_data/full`
- `/estimator/introspection_data/names`
- `/estimator/introspection_data/values`
- `/controller/introspection_data/full`
- `/controller/introspection_data/names`
- `/controller/introspection_data/values`
Then the behavior of the ROS parser is wrong as the static variable does not distinguish the different topics individually leading to a wrong parsing of the names versus data.

Therefore I created a map containing the prefix: `/controller/introspection_data` or `/estimator/introspection_data` where I store the topic names and versions.
Hence this map is filled up in the `parsePalStatisticsNames` and reused later depending on which topic the current message is read from.
Tried on Ubuntu22.04 humble.

### DIFF
--- a/plotjuggler_plugins/ParserROS/ros_parser.cpp
+++ b/plotjuggler_plugins/ParserROS/ros_parser.cpp
@@ -552,7 +552,8 @@ void ParserROS::parseDataTamerSnapshot(const std::string& prefix, double& timest
   DataTamerParser::ParseSnapshot(dt_schema, snapshot, callback);
 }
 
-static std::unordered_map<std::string, std::unordered_map<uint32_t, std::vector<std::string>>>
+static std::unordered_map<std::string,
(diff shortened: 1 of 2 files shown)


MaximilienNaveau:
Sorry I haven't seen: https://github.com/facontidavide/PlotJuggler/commit/077cc0080b85679337fcb3993f20c3762242f387
Solved with this commit

PR #1134, opened 2025-07-29
Add flake to build with nix

bigFin:
Flake to build PlotJuggler with Nix.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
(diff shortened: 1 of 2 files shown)


gvz:
hi, I gave your PR an try, as I was looking for exactly this. Thanks.
I found a few issues in it: 
1. it fails to load plugins, at least for me: 
```
Loading compatible plugins from directory:  "/nix/store/5cqvzmshb4813qbccyqzv6wbp7isj6m9-plotjuggler-3.10.11/bin"
"libdataload_mcap.so" :  "Failed to extract plugin meta data from '/nix/store/5cqvzmshb4813qbccyqzv6wbp7isj6m9-plotjuggler-3.10.11/bin/libdataload_mcap.so'"
"libDataLoadCSV.so" :  "Failed to extract plugin meta data from '/nix/store/5cqvzmshb4813qbccyqzv6wbp7isj6m9-plotjuggler-3.10.11/bin/libDataLoadCSV.so'" 
...
```
2. some plugins are not build at all, like DataLoadParquet and DataStreamMQTT

Not sure maybe the second one is intended as the github actions do not build these as well

gvz:
in #1140 I found a solution to build DataStreamMQTT

facontidavide:
already merging an alternative. I didn't have any preference between your version and the other, choose randomly.

If you believe that a change need to be applied, let me know. Thanks!!!!

PR #1143, opened 2025-08-04
Fix protobuf compilation on linux hosts

gvz:
This fixes a few issues that arise when compiling plotjuggler with protobuff support  on a linux host


facontidavide:
Thanks

Issue #816, opened 2023-04-22
Errors when launching AppImage


KronosTheLate:
# Problem description

When I download and launch the app-image, I get some error I am having a hard time parsing:
```
~ $MyPrograms/PlotJuggler-3.6.0-x86_64.AppImage 
qt.network.ssl: QSslSocket: cannot resolve EVP_PKEY_base_id
qt.network.ssl: QSslSocket: cannot resolve SSL_get_peer_certificate
qt.network.ssl: QSslSocket: cannot call unresolved function SSL_get_peer_certificate
Loading compatible plugins from directory:  "/tmp/.mount_PlotJuJtjNgc/usr/bin"
"libDataLoadCSV.so is a DataLoader plugin"
"libDataLoadMCAP.so is a DataLoader plugin"
"libDataLoadParquet.so is a DataLoader plugin"
"libDataLoadULog.so is a DataLoader plugin"
/tmp/.mount_PlotJuJtjNgc/AppRun.wrapped: symbol lookup error: /tmp/.mount_PlotJuJtjNgc/usr/bin/libDataStreamMQTT_Mosquitto.so: undefined symbol: mosquitto_lib_init
```

I could probably google around on how to fix the symbol lookup error. But the fix is not clear, and having to google around to simply launch the problem is a bit of a rough start. Also, I do not believe I have had time to do anything wrong yet, as I have only downloaded the AppImage, made it executable, and launched it. Which is why it seems like a problem with the program, which is why I open this issue.

# Steps to reproduce (important)
1. Attempt to launch app-image
2. 
- Describe your platform / Operative System.
<details>

```
~ $inxi -Fxzd
System:
  Kernel: 5.15.0-70-generic x86_64 bits: 64 compiler: gcc v: 11.3.0
    Desktop: Cinnamon 5.6.8 Distro: Linux Mint 21.1 Vera
    base: Ubuntu 22.04 jammy
Machine:
  Type: Laptop System: LENOVO product: 20NR0014MX v: ThinkPad L390
    serial: <superuser required>
  Mobo: LENOVO model: 20NR0014MX v: SDK0J40697 WIN
    serial: <superuser required> UEFI: LENOVO v: R10ET39W (1.24 )
    date: 08/12/2019
Battery:
  ID-1: BAT0 charge: 10.9 Wh (32.5%) condition: 33.5/45.0 Wh (74.5%)
    volts: 11.1 min: 11.1 model: LGC 01AV481 status: Discharging
[...]

PR #1146, opened 2025-08-14
build debian packages in github actions

gvz:
this pull request adds a job that builds .deb packages for:
- Debian 12
- Debian 13
- Ubuntu 24.04

I had the same problem as #816. After reading about OpenGL and Qt in AppImages I got the impression that it is not a good idea, so I started building .deb files


facontidavide:
thanks a lot for your help!

facontidavide:
The debians are correctly created but they are not uploaded to Release.

https://github.com/facontidavide/PlotJuggler/actions/runs/17359050438/job/49276406028

Any clue?

gvz:
yes, my bad. I did not rename the file before uploading it. I will test and send you a pull request

gvz:
this should work #1160

Issue #1136, opened 2025-07-30
Error loading ulogs on latest version 3.10.11


ischollETH:
When updating to the latest snap version 3.10.11 (on Ubuntu 22.04.), suddenly I was unable to open ulogs that I was able to before, with the attached error message. 
Downgrading back to 3.8.6 fixed the issue again. 
For privacy reasons I don't have a log at hand that showed this issue which I can share, but maybe others also have seen this and could add to this. 

<img width="359" height="133" alt="Image" src="https://github.com/user-attachments/assets/874d8395-a37f-4e0a-87e7-eeff798fb3da" />

bresch:
Here is a SITL log showing the issue:

https://logs.px4.io/plot_app?log=26ad54ed-8850-480f-b800-e2cca9059f35

Imti-git:
> When updating to the latest snap version 3.10.11 (on Ubuntu 22.04.), suddenly I was unable to open ulogs that I was able to before, with the attached error message. Downgrading back to 3.8.6 fixed the issue again. For privacy reasons I don't have a log at hand that showed this issue which I can share, but maybe others also have seen this and could add to this.
> 
[...]

PR #1154, opened 2025-08-28
fix(ulog_parser): allow messages without timestamp field

bresch:
fixes: https://github.com/facontidavide/PlotJuggler/issues/1136

The timestamp field is [optional](https://github.com/facontidavide/PlotJuggler/issues/1136#issuecomment-3191710284), so in case there is no timestamp, just a fixed increment.


facontidavide:
Thanks! I had no time to address this and I am grateful for your help.

5396edbd1d87ec2c7a041dffc4047e44

haumarco:
mvp

facontidavide:
@haumarco what do you mean by "mvp" in this context?

bresch:
I think he's happy that it got fixed :D ("Most Valuable Player")

Thanks for the quick response and merge @facontidavide !

bresch:
@facontidavide Could we get a new snap release soon, please, so other users can finally come back to a more recent version?

PR #1156, opened 2025-08-30
create new slider widget for adjusting line weight.

kitagawa444:
Create new slider widget in "Plot Widget Editor"for adjusting line weight.


facontidavide:
thank you for your contribution, but I am trying to avoid cluttetingthe UI with this kind of features

PR #1159, opened 2025-08-31
fix mqtt username and password swap

### LINKED ISSUE #1158: mqtt connection dialog swaps username and password on ever plotjuggler restart
Every time plotjuggler is restarted the username and password in the mqtt dialog are swapped, thus the password is shown on clear text every second time.

I will make a pull request with a fix later

gvz:
fixes  #1158


facontidavide:
thanks

PR #1173, opened 2025-09-20
Adjust color of cursor text to adhere to current theme

simonsag96:
At the moment, the color of the time: displayed with the new cursor it hardcoded to black.
This makes it hard to read when dark mode is enabled.

This PR changes this to that the text uses the Qt::WindowText Color instead of black.

image


facontidavide:
thanks a lot!

PR #1179, opened 2025-09-21
Fix Usability of MCAP Dialog

simonsag96:
If you currently load an MCAP, you get a dialog window where you can select the number when arrays should be truncated.
However, per default this spinbox for selecting the cursor is in focus (has a cursor in it).

This leads to the behavior that CTRL+A selects all the text inside the spin box instead of all topics in the bag.

image

This can be fixed by setting the spin box to click focus. 
This way, the number can still be edited by clicking into the spin box, but by default Ctrl+A selects all topics instead of the text inside the spin box.


facontidavide:
thanks!!!

PR #1180, opened 2025-09-25
[Feature] New Filter to display the time since the previous datapoint

simonsag96:
We often have the need to analyze the spacing of data points inside our (asynchronous) logs.
We do this to see if our software is properly executing in real-time or if there are certain problems in scheduling.

For this, we usually look at the time between consecutive messages.
Since we do that very often, i created a filter to directly compute the time difference between 2 data points.

Feel free to merge it. If you think that is not useful for most of the PJ users, it also fine to close this. Your call ;) 

image

### DIFF
--- a/plotjuggler_app/CMakeLists.txt
+++ b/plotjuggler_app/CMakeLists.txt
@@ -76,6 +76,7 @@ set(PLOtJUGGLER_SRC
     transforms/samples_count.cpp
     transforms/first_derivative.cpp
     transforms/scale_transform.cpp
+    transforms/time_since_last_datapoint.cpp
     utils.h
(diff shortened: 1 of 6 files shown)


ssys-zdavis:
I like the idea of a built-in c++ feature like this. We usually use lua to compute the frequency (hz) we're getting data. Maybe this filter could output both?

simonsag96:
You mean use the time since last point an calculate this into a frequency?

simonsag96:
Accidentally closed the PR, sorry for that.

ssys-zdavis:
> You mean use the time since last point and calculate this into a frequency?

Yes

PR #1181, opened 2025-09-27
[Feature] Add topic filtering into the mcap dialog

simonsag96:
This PR adds the possibility to filter for topics shown in the mcap dialog, very similar to how it is already possible inside the DataLoad ROS Plugin for a while now.
This is very handy when handling very big rosbags with hundreds of signals

Here is an image of the updated UI (notice the filter element on the right):

image


facontidavide:
thanks a lot !!!!!

PR #1185, opened 2025-09-29
[Bugfix] Add Tab Button being placed outside plot widget

simonsag96:
If you create a lot of tabs, the button to add a new one moves further and further to the right.
However, if you reach the width of the tabbed_plot_widget, the button just "falls" over the edge
of the widget to the right.
image

This PR fixes this by limiting the tab bar width to a maximum size, so button always remains there.

image


facontidavide:
Thanks a lot

image

PR #1192, opened 2025-10-03
Support curve multi-selection in Transform Editor

MichelJansson:
We often need to apply or remove the same transforms on many curves which has to be done one-by-one. This is tedious and error prone and multi-select makes intuitive sense, so this is my proposed solution.
Thanks!

Enhanced plotwidget transforms (Transform Editor) to support curve multi-selection, enabling batch operations of curve transforms.

- Updated curve list to handle multi-selection.
  - Changed curve selection mode to `ExtendedSelection`.
  - When multi-selection is started, the transform selection UI is cleared to make batch manipulation explicit (ie. batch transform not applied until explicitly selected).
- Changed transform list for multi-curve updates.
  - Transform state is applied to all selected curves using existing XML serialization api.
  - Alias logic updated to better stay in sync with currently selected transform unless manually changed by checking it the current alias is the default alias for the previous transform.

40fb11ba-a27c-4600-9d59-914b7409001d

### DIFF
--- a/plotjuggler_app/plotwidget_transforms.cpp
+++ b/plotjuggler_app/plotwidget_transforms.cpp
@@ -107,18 +107,24 @@ QColor DialogTransformEditor::RowWidget::color() const
 
 void DialogTransformEditor::on_listCurves_itemSelectionChanged()
 {
-  auto selected = ui->listCurves->selectedItems();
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks a lot, I had this on my TODO list... 

image

PR #1201, opened 2025-10-24
Introduce pixi for dependency management and build orchestration

tdejager:
Hey @facontidavide, this is one of the pixi maintainers here. This is my first attempt at using PlotJuggler as a use-case to build with RoboStack and standalone using pixi and the pixi build. Feature, more on this: https://pixi.sh/latest/ and specifically pixi build: https://pixi.sh/latest/build/ros/ and https://pixi.sh/latest/build/cpp/.

This gives users the ability to build and run PlotJuggler using a couple of different configurations, with just a pixi installation, no cmake or ROS needed. There are still some drawbacks, like the ROS plugins not being available for noetic on anything but linux-64, and for jazzy just missing on windows. As well as the project rebuilding too often, and needing a kind of weird configruration of multiple pixi.toml's for the different package configurations.

```
  PlotJuggler/
  ├── pixi.toml                    # Main workspace (defines 3 environments)
  │                                # - jazzy (default, ROS2)
  │                                # - noetic (ROS1, linux-64 only)
  │                                # - non-ros
  │
  ├── pixi_jazzy/
  │   └── pixi.toml                # ROS2 Jazzy build config
  │                                # Uses: pixi-build-ros backend
  │
  ├── pixi_noetic/
  │   └── pixi.toml                # ROS1 Noetic build config
  │                                # Uses: pixi-build-ros backend
  │
  └── pixi_non_ros/
      └── pixi.toml                # Standalone build config
                                   # Uses: pixi-build-cmake backend
                                   # Declares all deps: Qt, Lua, fmt, lz4, zstd, etc.
```
Note the ros backends use the package.xml for reading dependency information, while the cmake version has the dependencies specified.

However, I do feel its a promising start, so that's why I'm opening the PR in draft so you can take a quick look. To test it out you need a pixi installation: https://pixi.sh/latest/installation/

Some things I needed to do to get things compiling, I've mainly tested on macOS arm.

### Main CMakeLists.txt
  - New options: PREFER_DYNAMIC_ZSTD and PREFER_DYNAMIC_LZ4 to control library linking preference
  - Fixed linking: Added Qt5::Xml to plotjuggler_base target (was missing)
### LZ4 detection (cmake/find_or_download_lz4.cmake) &  ZSTD detection (cmake/find_or_download_zstd.cmake)
  - Respects PREFER_DYNAMIC_LZ4 flag, needed this otherwise it would use the static library from my system.
  - Update target detection: Checks for existing LZ4 targets from system packages or toolchains (multiple naming conventions: LZ4::lz4, lz4::lz4, lz4, etc.)
  - Fallback with preference: If both shared and static exist, chooses based on PREFER_DYNAMIC_LZ4
  - ParserProtobuf:
    - Added missing absl_log_internal_message library for macOS

## Pixi 

To run with pixi there is basically a `pixi run start` task for the different environments and that should build and install it correctly into an isolated environment. Take a look at the `pixi.toml` to kind of get the gist.

## Future work
- There is a lua code-signing on OSX issue that needs a new pixi release (we've fixed this in main)
- Need to test the ROS configurations on windows and linux better. And at least get a jazzy/rolling version of the ROS plotjuggler plugins on Windows.
- I could set-up CI for this, but first wanted to see if this change is interesting at all.
- We have a development feature that makes developing with the environment in the IDE easier: https://github.com/prefix-dev/pixi/issues/4721. This will make this whole thing infinitely more useful I think.

Anyways, sorry for the long description, and would be great to hear your thoughts.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -10,16 +10,16 @@ environments:
       - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
       - conda: https://prefix.dev/conda-forge/linux-64/alsa-lib-1.2.14-hb9d3cd8_0.conda
-      - conda: https://prefix.dev/conda-forge/noarch/argcomplete-3.6.2-pyhd8ed1ab_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/argcomplete-3.6.3-pyhd8ed1ab_0.conda
       - conda: https://prefix.dev/conda-forge/linux-64/attr-2.5.2-h39aace5_0.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/binutils-2.44-h4852527_2.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/binutils_impl_linux-64-2.44-hdf8817f_2.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils-2.44-h4852527_4.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils_impl_linux-64-2.44-h9d8b0ac_4.conda
       - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
       - conda: https://prefix.dev/conda-forge/linux-64/c-ares-1.34.5-hb9d3cd8_0.conda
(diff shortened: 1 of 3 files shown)


tdejager:
Noticed some annoying things, with this still upstreaming some work to pixi to fix this :)

facontidavide:
any progress? Looking forward to see this in action

tdejager:
Next release should fix some of the caching issues on pixi side of things, and then I'll ping you for a test drive :)

facontidavide:
I will close this until the final solution is proposed. looking forward to see this merged :smiley:

PR #1205, opened 2025-10-31
Added a script to remove orphaned link upon uninstall on debian

lakid:
The symbolic link `/usr/bin/plotjuggler`  gets created on postinst, but is left behind upon package removal.
This was also causing reinstallation to display an error stating that the link already exists (i.e. returns non-zero).

This change adds a postrm script to the Debian package to remove the orphaned symbolic link.


gvz:
good catch, does this also fix the error message while updating?
anyways would be good to merge this

lakid:
Yes, also fixes the error message while updating, for updates from this change and after to newer.
Updates from an old version to this will still have that error message. 
It can also be fixed, if needed.

PR #1209, opened 2025-11-19
Parquet loading

gvz:
This adds the capability to read the timestamp data type from parquet files with the correct conversion to PlotJuggler's internal time.
It also fixes a little bug, where bool columns were shown in the selector but were not displayed when pulled into the plot. Now they are plotted as 1 and 0.


facontidavide:
thanks

facontidavide:
image

PR #1210, opened 2025-11-19
UI additions

gvz:
This add quiet some little quality of live UI addition. If you want some or all in separate PRs let me now.
 - add toggle button to turn all data series into stepped lines
 - add toggle button to show dots at samples, it also makes dots a little larger so they are actually bigger than lines
 - add toggle button that converts all timestamps to UTC zone
 - add toggle button to display the x value in the pointer display in ISO date format

image

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -291,6 +291,11 @@ MainWindow::MainWindow(const QCommandLineParser& commandline_parser, QWidget* pa
   ui->buttonUseUtc->setChecked(_use_utc_time);
   connect(ui->buttonUseUtc, &QPushButton::toggled, this, &MainWindow::on_buttonUseUtc_toggled);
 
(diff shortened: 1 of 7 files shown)


Bartimaeus-:
Got any pics of your UI changes?

gvz:
sorry, should have included it directly. 
In this screenshot all the things are enabled
image

facontidavide:
I feel like this is trying to do too much and I don't really like it from a cosmetic point of view.

I will cherry pick the "display points" only, since I believe it is very useful

gvz:
Do you see any low hanging fruits which I could change and would convince you to cherry pick more?
Would you prefer to have buttons like the UTC one as something the user sets in the settings of PlotJuggler?

facontidavide:
Just pushed this:

https://github.com/facontidavide/PlotJuggler/commit/c8ab24b964c9854e4f1119723bd513aaa329dcc2

Issue #1218, opened 2025-12-02
Export Data to CSV Directly from Plot


lars-nagel:
## Summary
Add the ability to directly export time-series data to CSV from individual plots, respecting the current zoom level and selected data series.

## Description
Currently, while the application offers a "Publishers -> CSV Exporter" feature and options to "Save plot to file" or "Show data statistics" (with "Calculate statistics from: Visible Range"), there is no direct way to export the raw data from a specific plot for only the currently visible/zoomed time range.

## Proposed Feature
Add a context menu option (accessible via right-click on a plot) called **"Save Data to CSV"** or **"Export Data to CSV"** that would:

- Export only the time-series data currently displayed in the selected plot
- Respect the current zoom level (only export data within the visible time range)
- Provide a flexible, per-plot export workflow

## Use Case / Motivation
This feature would enable users to quickly extract specific subsets of data from their plots without having to:

- Export all data and manually filter it
- Lose the context of their current visualization state

This is particularly useful for:

- **Sharing specific time windows** of data with colleagues
- **Further analysis** of zoomed regions in external tools
- **Documentation and reporting** purposes

lars-nagel:
In any case, thank you very much for developing this great tool!

lars-nagel:
I prototyped how I imagine this feature here: #1219.

facontidavide:
Thanks for the prototype. the idea is great, but as I mentioned in the PR, I have a better plan that includes a new "Export toolbox" that will include CSV among other formats.

I am closing this as "wontfix", but thatnks a lot for the idea!!

PR #1219, opened 2025-12-03
Add 'Export Data to CSV' option to plot context menu

lars-nagel:
## Summary
- Adds right-click context menu option "Export Data to CSV" to plots
- Exports only data within the visible/zoomed time range
- Supports multiple curves merged by timestamp

## Motivation
Addresses the feature request for per-plot CSV export respecting zoom level (see issue #1218).

Currently users must use the global CSV Publisher to export data all loaded data. This adds a quick way to export just the visible portion of a specific plot.

### DIFF
--- a/plotjuggler_app/plotwidget.cpp
+++ b/plotjuggler_app/plotwidget.cpp
@@ -11,6 +11,8 @@
 #include <QDrag>
 #include <QDragEnterEvent>
 #include <QDragMoveEvent>
+#include <QDir>
+#include <QFile>
 #include <QFileDialog>
 #include <QFileInfo>
 #include <QMessageBox>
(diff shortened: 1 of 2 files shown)


facontidavide:
Before you work more on this, please take into account that most probably I will not merge it in.

I am planning a complete refactoring of the CSV export to be much better than it currently is, therefore this change may not fit in my plan

facontidavide:
Thanks a lot, but as I said, I will do this differenlty.

I appreciate your contribution and I promise I will give you something even better than this in the next few months

lars-nagel:
Haven't followed up on this for quite a while.
Amazing to hear about the new features that you have planned there - I am looking forward to it!

PR #1234, opened 2025-12-19
Fix: Explicitly use UTF-8 encoding for layout load/save 

### LINKED ISSUE #1221: Layout Import/Export compatibility with special characters
Heyo,

I've encountered an issue when exporting a layout with a plot that utilises unicode characters, in this case, the ° character for degrees.

PlotJuggler seems to parse these lines just fine, and lets you export a layout w/ data, but upon starting a new session and importing the layout it fails to find the same datapoints.

This is on Win64 V3.14.1

Image

Pana1v:
Fixed the issue by specifying UTF-8 The Issue did not occur on Linux, however did on Windows.

closes #1221

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -1901,17 +1901,24 @@ bool MainWindow::loadLayoutFromFile(QString filename)
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


Pana1v:
@facontidavide PTAL

PR #1235, opened 2025-12-20
Feature: add option to swap pan/zoom mouse actions

viktor-av:
This adds button to the side panel to optionally swap pan/zoom mouse actions. Also includes changes in cursor icon setting logic.


facontidavide:
Thanks, but no thanks.

At most, I can accept a change in the "Preference" dialog, but not a new button in the main UI

viktor-av:
ok, fair point
I have changed and rebased the branch accordingly

PR #1244, opened 2025-12-25
Fix locale-dependent decimal parsing in CSV reader

### LINKED ISSUE #1237: [BUG]: Decimal separator not detected properly
**Describe the bug**
After the refactoring of the CSV reader, PlotJuggler fails to parse decimal numbers properly. It always uses the decimal separator from the locale.

**To Reproduce**
1. Start PlotJuggler with a locale that uses comma as decimal separator, for example `LANG=de_DE.UTF-8 plotjuggler`
2. Open [test.csv](https://github.com/user-attachments/files/24327880/test.csv), that uses a period as a separator.

**Expected behavior**
Decimal numbers are parsed as decimal numbers (e.g. `20.66`), instead they are truncated to the integer part (`20`).

**System details**
Version 3.14.4

**Additional context**
Works with a matching locale, e.g. `LANG=en_US.UTF-8 plotjuggler`.

Pana1v:
Replace std::stod() with locale-independent std::from_chars() to ensure decimal numbers are parsed correctly regardless of system locale settings.

Fixes #1237


facontidavide:
doesn't compile for macos!

facontidavide:
I found a different solution using Qt provided function, that I think is better for portability.

FYI:

```
std::optional<double> toDouble(const std::string& str)
{
  // If comma is used as decimal separator (European format), replace with dot
  QString qstr = QString::fromStdString(str);
  qstr.replace(',', '.');

  // Use QLocale::c() for locale-independent parsing (always uses '.' as decimal separator)
  bool ok = false;
  double value = QLocale::c().toDouble(qstr, &ok);
  return ok ? std::optional<double>(value) : std::nullopt;
}
```


thanks a lot anyway for looking into this

Issue #1126, opened 2025-07-22
ZMQ Subscriber Configuration Lacks Option to Specify Protobuf Paths


aerobastian:
When setting up a ZMQ subscriber to receive Protobuf messages, you have options labeled "Proto File" and "Include Folders" to specify the Protobuf file or directories. However, there is no visible button or input field to add or specify the paths. See the attached image for reference.

<img width="543" height="429" alt="Image" src="https://github.com/user-attachments/assets/f1f52c16-903a-4ac3-8727-ca9bd55d8ada" />

<img width="543" height="429" alt="Image" src="https://github.com/user-attachments/assets/5d08e947-de1a-44aa-afa5-cfd9aa92f6c6" />

aerobastian:
Would really appreciate some help with this.

bpousett:
@aerobastian Try dragging the menu window down with the mouse to expand the options. This is what it should look like:

<img width="557" height="838" alt="Image" src="https://github.com/user-attachments/assets/218e7bcc-214b-433d-acd1-2ec321e794fe" />

I also was confused at first. Perhaps the menu window is too small by default.

facontidavide:
In PlotJuggler 4 the ZMQ dialog no longer embeds the (clippable) protobuf fields: you pick the parser encoding in the ZMQ dialog, and the protobuf parser has its own dialog with explicit 'Load .proto file' and 'Add include folder' buttons. Closing as resolved by the PJ4 rewrite.

---
*PlotJuggler 4 is now the official release and the default branch of this repo; PlotJuggler 3 is end-of-life and no longer receives features or fixes. If this still affects you in PJ4, please reopen or file a fresh issue — it will be looked at.*

PR #1254, opened 2026-01-05
Adjust dialog size based on protocol selection in DataStreamZMQ

sachinkum0009:
Modify the dialog's size dynamically based on the selected protocol to enhance user experience.

It targets the issue #1126

Please LMK if you have some suggestions. Thanks


facontidavide:
I can see that it is still in draft. let me know when it is ready for final review. looks good so far, but I haven't tested it

Issue #1220, opened 2025-12-06
Option to incorporate dates & times that are stored in different columns


Blabzillaweasel:
As per picture below current PlotJuggler design expects date and time in the same cell _(I imagine this is fairly standard)_.

Some logging equipment I work with stores this information in seperate columns, being able to specify date in one column and time in another column and have PlotJuggler combine the info would be neat.

Example image attached.
<img width="1178" height="711" alt="Image" src="https://github.com/user-attachments/assets/b4d8d12a-5bbb-4b58-855b-89e4ed9263fe" />

facontidavide:
PR welcome

facontidavide:
honestly, I think this is uncommon and a little TOO specific for your use case.

Can you be more specific about the equipment you use, to understand if that is really that common?

Blabzillaweasel:
That's entirely understandable. These logs specifically are from Cummins engine control modules used in various heavy diesel machines, trucks, buses etc.

MrSerafim:
I'm actually having the same problem. I use PlotJuggler to visualize CSV flight logs from RC models, and EdgeTX (almost every transmitter uses this OS) stores date and time in separate columns. So I think this feature would be great; 
A lot of RC pilots (RC planes, FPV drones, RC cars) use EdgeTX with this same format. And right now there really isn't a good tool to analyze logs from this controller, most people use Excel, so I think it would be very nice to have this in PlotJuggler.

facontidavide:
PlotJuggler 4's CSV loader supports this directly via the 'Combine Date + Time columns' option, which merges a date-only column and a time-only column (with auto-detection of adjacent pairs) into a single timestamp — exactly the EdgeTX/Cummins layout described here.

---
*PlotJuggler 4 is now the official release and the default branch of this repo; PlotJuggler 3 is end-of-life and no longer receives features or fixes. If this still affects you in PJ4, please reopen or file a fresh issue — it will be looked at.*

PR #1259, opened 2026-01-07
Add automatic detection and merging of separate date/time column (#1220)

lucabonamini:
**Summary**

Added automatic detection of CSV files with date and time stored in separate columns, with automatic combination into a single timestamp for plotting.

**Motivation**

Some data logging equipment stores date and time information in separated CSV columns:
```
Date,Time,Temperature
2026-01-01,10:30:25.000,23.5
2026-01-01,10:30:26.000,23.6
```

PlotJuggler previously required timestamps in a single column.

**Solution**

Implemented automatic detection and merging:

1. New column types: Added DATE_ONLY and TIME_ONLY to ColumnType enum
2. Detection logic: Automatically identifies date-only columns (e.g., "2026-01-01") and time-only columns (e.g., "10:30:25.000")
3. Pairing algorithm: Finds adjacent date/time column pairs
4. Virtual columns: Creates combined "Date + Time" virtual columns in the UI
5. Timestamp generation: Parses and combines separate date/time values into Unix epoch timestamps with fractional second support 

---

**Timeline slider fix**

The timeline slider was diving the time range by the number of data points instead of the number of intervals between them, causing navigation steps to display incorrect time values. This fix ensures the slider accurately represents the actual time spacing in the data.

---

**Fix tooltip display at end of time series data**

When the cursor is positioned at or beyond the last data point in a time series, the tooltip now correctly shows the final data point value instead of disappearing. This ensures consistent tooltip behavior across the entire data range, including at the boundaries.

### DIFF
--- a/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
+++ b/plotjuggler_plugins/DataLoadCSV/dataload_csv.cpp
@@ -215,9 +215,17 @@ DataLoadCSV::DataLoadCSV()
     bool box_enabled = !checked || selected.size() == 1;
     _ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(box_enabled);
   });
+  connect(_ui->radioButtonDateTimeColumns, &QRadioButton::toggled, this, [this](bool checked) {
+    _ui->listWidgetSeries->setEnabled(!checked && _ui->radioButtonSelect->isChecked());
+    if (checked)
+    {
+      _ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(true);
+    }
+  });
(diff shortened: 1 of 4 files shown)


lucabonamini:
Related to #1220

facontidavide:
I wonder if we shouldn't add a radio option for this case instead of relying on magically finding the pair and adding it to the columns.

image

lucabonamini:
> I wonder if we shouldn't add a radio option for this case instead of relying on magically finding the pair and adding it to the columns.
> 
> image

Done. I've replaced the automatic list widget item with a dedicated radio button option. When adjacent date+time column pairs are detected, the radio button becomes enabled and shows which columns will be combined.

facontidavide:
since the original PR has now a lot of conflict, i am cherry picking, but I could not do this without your input. thanks a lot

image

Issue #1207, opened 2025-11-06
State of legend and tracking info not synced between subplots


SammysHP:
If you split the view, the new subplot does not inherit the current state of the legend or tracking info, but uses the default (right/off).

## Steps to reproduce

1. Set legend to left and enable cursor tracking info.
2. Split view.

## Actual result

The new subplot has the legend on the right side and no cursor tracking info.

Pressing the corresponding buttons advances the mode (no legend, more info).

## Expected result

Same state as in other subplot.

## Additional information

PlotJuggler 3.13.2

facontidavide:
thanks for reporting!!!

PR #1260, opened 2026-01-07
Inherit legend settings when splitting plot views

lucabonamini:
**Summary**

Fixed an issue where newly created subplots (via split view) did not inherit the current legend settings from existing plots, causing them to always display the default legend configuration (right-aligned, visible) instead of matching the user's current preferences.

**Motivation**

When users configure their legend settings (left/right alignment, hidden/visible) and then split a view to create additional subplots, they expect the new subplot to maintain the same visual configuration as the existing plots. However, the new subplots were always created with default settings:

- Legend always positioned on the right side
- Legend always visible (even if user had hidden it)
- Pressing legend buttons would advance through modes instead of matching current state

This inconsistent behavior disrupted the user's workflow and required manual reconfiguration of each new subplot to match their preferences.

**Solution**

Enhanced the `onPlotAdded` method in `mainwindow.cpp` to inherit legend settings from the application's current state:

- New subplots now inherit legend visibility from `_labels_status` (hidden/visible)
- New subplots now inherit legend alignment from `_labels_status` (left/right positioning)
- Legend configuration is applied immediately when the subplot is created

The fix ensures that split views maintain visual consistency across all subplots, matching user expectations and improving the overall user experience.

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -832,6 +832,17 @@ void MainWindow::onPlotAdded(PlotWidget* plot)
   plot->setDefaultStyle(ui->buttonDots->isChecked() ? PlotWidgetBase::LINES_AND_DOTS :
                                                       PlotWidgetBase::LINES);
 
+  // Inherit legend settings from current state
+  plot->activateLegend(_labels_status != LabelStatus::HIDDEN);
+  if (_labels_status == LabelStatus::LEFT)
+  {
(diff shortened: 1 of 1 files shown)


lucabonamini:
Related to #1207

facontidavide:
careful, I think you included changes from #1259 (CSV)

lucabonamini:
Apologies for the mistake, I had accidentally included files from a different PR. This has now been fixed.

facontidavide:
thanks a lot!

Issue #1015, opened 2024-10-05
UDP Server crashing


Lsnso:
Hi.

I am running PopOS 20.04 with the latest main commit 591048.
When I try to set up a UDP Server, my window crashes.

Below is a video showing the behaviour. Dummy streamer works just fine.

https://github.com/user-attachments/assets/02735a11-020c-4326-bd86-9d32becad83e

Thanks in advance!

Petapton:
Hello, I get this same error too. I'm using AUR-built version 3.9.2.
WebSocket Server and ZMQ Subscriber don't seem to be affected. Settings window opens correctly for them.

Ekstrapolator:
check on witch branch you are, and if not on master switch and rebuild plotjuggler and udp plugin

tarunkartik:
Can I take this issue? Happy to submit a PR.

facontidavide:
be my guest!

PR #1265, opened 2026-01-14
Fixes issue #1015: UDP server crash

tarunkartik:
### Summary
Fixes issue #1015  where PlotJuggler crashes when trying to start a UDP server.
- Was not able to recreate the exact scenario that reporter experienced but the stack trace in the video shows that the error was an std::out_of_range exception within UDP::start().
- The only possible location that can happen is when the at() function is called on the parserFactories()

### Changes
- Replaced the  call to at() with a safer lookup by calling find() inside the lambda for onComboChanged()
- Provided a fallback in case the selected_protocol does not exist

### Testing
- Built in docker container
- Manually tested UDP streaming

### DIFF
--- a/plotjuggler_plugins/DataStreamUDP/udp_server.cpp
+++ b/plotjuggler_plugins/DataStreamUDP/udp_server.cpp
@@ -118,7 +118,20 @@ bool UDP_Server::start(QStringList*)
         prev_widget->setVisible(false);
       }
     }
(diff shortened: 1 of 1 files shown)


facontidavide:
nice ,but I don't thing that we should default to another one. we should just close the connection, IMO

facontidavide:
fixed differently, thanks

Issue #1255, opened 2026-01-06
[BUG]: Line Styles Cannot be Changed Once Set


Bartimaeus-:
**Describe the bug**
Line style cannot be changed once modified from the Edit Curves dialog window

**To Reproduce**
Add curves to plot
Edit line style with the Edit Curves window
Save changes
Attempt to change the line style again

See video

https://github.com/user-attachments/assets/76a50f47-bc65-4290-a2c1-6f0998776051

**Expected behavior**
Changes to line style applied after clicking on "save changes"

**System details**

 - OS: Windows 11
- Platform: N/A
 - Version 3.15.0

Pana1v:
This can be closed

PR #1277, opened 2026-01-23
Fix Reset to Default Curve Style in PlotWidget

Pana1v:
Ensures plot curve styles correctly revert to "Lines (default)" when loading layouts by explicitly clearing overrides when the style attribute is missing.
Fixes Issue #1255 
## Before
[Screencast from 2026-01-23 22-50-51.webm](https://github.com/user-attachments/assets/085df066-1513-49c8-99b6-4601e9cbaa6d)
## After
[Screencast from 2026-01-23 22-51-54.webm](https://github.com/user-attachments/assets/36bad35a-b964-44ae-b740-7c0dd4e6c616)


Pana1v:
PTAL @facontidavide

PR #1278, opened 2026-01-24
Fix: Save colorMaps section to layout XML

Pana1v:
Fixes #1258
The colorMaps element was being created and populated but never appended to the XML document root, causing custom colormaps to be lost when saving layouts.


Pana1v:
Ptal @facontidavide

facontidavide:
image

PR #1289, opened 2026-02-12
fix: quaternion-to-RPY toolbox streaming data loss

### LINKED ISSUE #1059: Quaternion to RPY toolbox streaming only displays last point
When setting up the roll, pitch, yaw with the toolbox then it works well until I start the paused stream (from ROS2) as can be seen in the image below. It looks like only the last point is converted and saved. 

Pressing pause, using the toolbox and looking at the data:

Image

Presssing play to continue the stream:

Image

The manual custom series works as expected (but is also handled different internally as far as I know).

pradhankukiran:
- Remove `data_roll.clear()`, `data_pitch.clear()`, `data_yaw.clear()` calls from `QuaternionToRollPitchYaw::calculate()` that wiped output every replot cycle (40ms), leaving only 1-2 points visible during live streaming
- Change timestamp guard from `>=` to `>` to prevent duplicate point insertion on each cycle

The `calculate()` method now follows the same incremental append pattern used by `TransformFunction_SISO` and all other streaming transforms. External code in `TransformedTimeseries::updateCache()` already handles clearing + reset when a full recompute is needed.

  Fixes #1059

### DIFF
--- a/plotjuggler_plugins/ToolboxQuaternion/quaternion_to_rpy.cpp
+++ b/plotjuggler_plugins/ToolboxQuaternion/quaternion_to_rpy.cpp
@@ -33,10 +33,6 @@ void QuaternionToRollPitchYaw::calculate()
   data_pitch.setMaximumRangeX(data_x.maximumRangeX());
   data_yaw.setMaximumRangeX(data_x.maximumRangeX());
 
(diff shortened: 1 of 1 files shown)


facontidavide:
image

PR #1290, opened 2026-02-12
fix: CSV export missing values for quaternion and RPY series

### LINKED ISSUE #1137: Quaternion values as well as roll/pitch/yaw are not exported to the CSV file
I think this might be a bug, unless i am missing something. I have a topic of imu data and when i try to export it to CSV, the columns are correctly transferred, but there are no values inside them.

pradhankukiran:
Fixes #1137  - quaternion and roll/pitch/yaw columns appear in CSV export with headers but no values.
Four root causes identified and fixed:

  1. **Off-by-one in index initialization** (`publisher_csv.cpp:291`): `indices[i] = index + 1` skipped the first in-range sample of every series. Changed to `indices[i] = index`, consistent with `generateStatisticsCSV()`.

  2. **Machine-epsilon timestamp matching** (`publisher_csv.cpp:328`): Values from different series were merged into a CSV row only when timestamps matched within `std::numeric_limits<double>::epsilon()` (~2.2e-16). Real sensor data has larger floating-point drift between channels, causing cells to go blank. Replaced with an adaptive tolerance (`min_dt * 0.5`) derived from the minimum sample period across all exported series — mathematically guaranteed to never merge distinct adjacent samples.

  3. **NaN values stalling index advancement** (`publisher_csv.cpp:344-348`): The index only advanced when a value was not NaN, making it impossible to distinguish "no data at this timestamp" from "data is NaN at this timestamp." Added a `row_used` boolean vector to track participation explicitly, so indices advance correctly regardless of
  NaN status.

  4. **No zero-norm quaternion guard** (`special_messages.cpp`, `quaternion_to_rpy.cpp`): A zero quaternion `(0,0,0,0)` caused `1.0 / sqrt(0) = inf`, producing NaN for all RPY values, which then triggered 3rd bug. Added an early return of `(0,0,0)` RPY when `quat_norm2 < 1e-10`.


facontidavide:
CI is red sir

facontidavide:
nevermind, that was my fault. for the PR, actually you should return NaN if you want plotJuggler to "skip it". do not return 0.0

pradhankukiran:
@facontidavide done, returning NaN instead of 0.0 now. thanks for finding it out.

facontidavide:
this plugin has been retired

PR #1297, opened 2026-02-18
Fix Parsing of DiagnosticArrays

simonsag96:
At the moment, when parsing a diagnostic_msgs/DiagnosticArray, only the key value pairs inside each diagnostic status are pushed into the map of available signals.

Therefore, you cannot see the actual error level and message of the diagnostic status itself.

With this change, the message, the level and the key values pairs are correctly parsed.

Before:
image

After:
image


simonsag96:
Hm, the pipeline fails with `no module named ament_package'.

I am pretty sure my changes should not cause this. 
Maybe ROS2 is not sourced on the CI runner @facontidavide ?
But i am not sure why this would occur now...

facontidavide:
thanks!

Issue #1279, opened 2026-01-25
[BUG]: Values of string fields fail to display in timeseries list if over 15 characters


wojciech-graj:
**Describe the bug**
In the timeseries list, string values of length >15 characters do not display correctly.

**To Reproduce**
Import a csv file with the following contents, indicating that the time column contains timestamps ([test.csv](https://github.com/user-attachments/files/24848706/test.csv)).
```csv
time,x
0.0,Test123Test123T
1.0,Test123Test123Te
```

Scrub to the 0.0 timestamp and observe that the value of column x is displayed correctly as "Test123Test123T" in the Timeseries List (left middle of the screen).

<img width="994" height="665" alt="Image" src="https://github.com/user-attachments/assets/de2eb10c-1a63-45b3-b9dd-1ad3500d4d83" />

Scrub to the 1.0 or 0.5 timestamp and observe that the value of column x is displayed incorrectly in the Timeseries List.

<img width="994" height="665" alt="Image" src="https://github.com/user-attachments/assets/2b8fd4e7-4f45-4384-b2c2-eb7808c5fe0a" />

**Expected behavior**
I expect the string to display in the same way that it did when it was 15 characters long, when it is over 15 characters long. 

**System details**

 - OS: Debian GNU/Linux 13 (trixie)
- Platform: Appimage
 - Version 3.15.0

**Additional context**
The issue is caused by use-after-free in a `StringRef`. The `StringRef` owns the string if it is <= 15 characters, hence why the issue is only present with longer strings. I haven't been able to create a patch that fixes this.

simonsag96:
I can also reproduce the issue, with parsing from ros2.

The issue was introduced in commit 1ffc59896579a34c86d640109101082d983a9a2a

simonsag96:
I tried fixing this, but I am really not sure if I didn't accidentally break something else with it.

PR #1298, opened 2026-02-18
Fix merging for using string series

simonsag96:
This addresses #1279

However, i am not sure if i actually got the logic right for the string overload to work in all situations.
@facontidavide  take this with caution, it fixes the issue for me, but i don't kniw if its the best solution to the problem.


facontidavide:
image

PR #1299, opened 2026-02-18
feat: add support for plotting the last array element with [-1]

bektaskemal:
I sometimes need to plot the last element of a dynamic-size array, but PlotJuggler only exposes fixed indices like [0], [1], … and the tree keeps showing indices up to the maximum size ever seen. So when the array shrinks/grows over time, you can’t easily know which index is actually “the last one right now” to plot.

This MR adds virtual [-1] series aliases so you can plot things like poses[-1] or poses[-1]/position/x and always get the last parsed element at each sample (including nested arrays). [-1] is grouped under [] in the tree like other indices.

Also fixes a small protobuf clamp bug (std::max → std::min) [here](https://github.com/facontidavide/PlotJuggler/pull/1299/changes#diff-d67f8d176b1f6e96575613c43f2cdae5b801bfc627940d182f296e53b748f812R118) so truncation works as intended.

#### Screenshots:
back_alias_tree_view
back_alias_plot

### DIFF
--- a/plotjuggler_app/curvelist_panel.cpp
+++ b/plotjuggler_app/curvelist_panel.cpp
@@ -376,7 +376,7 @@ void CurveListPanel::refreshValues()
 
 QString StringifyArray(QString str)
 {
-  static const QRegExp rx("(\\[\\d+\\])");
+  static const QRegExp rx("(\\[-?\\d+\\])");
   int pos = 0;
   std::vector<std::pair<int, int>> index_positions;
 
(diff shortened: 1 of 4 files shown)


facontidavide:
I am not sure I like this...

bektaskemal:
Why? Do you think the use case is not valid, or is the concern more about the implementation?

maracuya-robotics:
What if I need the second to last item? It's not great to duplicate topics in the list, just with different names. What we need instead is a way to dynamically reference topics, like accessing topics in custom timeseries by their name instead of v1/v2/… generated from the fixed input list.

facontidavide:
I agree with @maracuya-robotics

bektaskemal:
Makes sense, thanks.

PR #1300, opened 2026-02-19
feat: implement topological sorting for nested dependencies in custom equations

tstienstra:
It is awesome that one can easily define custom equations in PlotJuggler. However, I was running into the problem that I was getting errors when having dependencies between equations. This PR solves this problem by introducing a topological sort algorithm (Khan's algorithm) to sort the snippets.


facontidavide:
very nice, thanks!

PR #1301, opened 2026-02-19
fix: parsing of nested topics in ulogs

### LINKED ISSUE #1272: [BUG]: Timestamp of position_setpoint_triplet parsed incorrectly
**Describe the bug**
It seems that the `position_setpoint_triplet` (nested ULog message) isn't parsed entirely correct. As documented [here](https://docs.px4.io/main/en/msg_docs/PositionSetpointTriplet), the timestamp is defined in the "parent" format. I suspect that the recursive solution that came out of #151 doesn't take it into account.

**To Reproduce**
Load a ulog file and plot some of the fields from `position_setpoint_triplet`, e.g. `position_setpoint_triplet/current/lon` and check the time axis.

**Expected behavior**
I would expect the timestamps defined in the parent format to be passed down to children.

**System details**

 - OS: Ubuntu 24 LTS
- Platform: snap
 - Version 3.15.0

- PX4 version: v1.16.0

**Additional context**

tstienstra:
This PR fixes #1272

### DIFF
--- a/plotjuggler_plugins/DataLoadULog/ulog_parser.cpp
+++ b/plotjuggler_plugins/DataLoadULog/ulog_parser.cpp
@@ -157,7 +157,7 @@ void ULogParser::parseDataMessage(const ULogParser::Subscription& sub, char* mes
 }
 
 char* ULogParser::parseSimpleDataMessage(Timeseries& timeseries, const Format* format,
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks!

PR #1310, opened 2026-02-27
fix: timestamp detection with array fields

tobias-auterion:
Fixes a crash when loading ulog files where the timestamp field is preceded by array fields.

The old code compared a cumulative data index against a field index to find the timestamp. These only match when all preceding fields are scalars. Any array field before the timestamp causes them to diverge, either misreading data as the timestamp or skipping it entirely and crashing on out-of-bounds access.

Timestamp bytes are now consumed at the correct stream position based on field index. Also fixes nested type handling which previously assumed timestamps were always first, and hardens format parsing against extra whitespace.


facontidavide:
thanks a lot

Issue #1232, opened 2025-12-19
[Feature]: Add serial port data streaming.


Valle125:
Hi, 

I'd like to add a serial port plugin to plotjuggler. I know that serial port data can be displayed in plotjuggler e.g. via a UDP bridge. However this poses some hurdle to less tech savvy people. I've observed people not wanting to use plotjuggler because of the missing out of the box support for serial ports. That's why I think it would be a great addition to have a serial port plugin.

I've some ideas about how the plugin could look and what features it shall have. Basic serial port configuration should be possible (see picture).
<img width="902" height="257" alt="Image" src="https://github.com/user-attachments/assets/c3921a3c-2562-4277-96ac-dc180b783670" />

ASCII (and later maybe binary) data shall be supported and the format shall be configurable (see picture).
<img width="902" height="257" alt="Image" src="https://github.com/user-attachments/assets/e41b2826-9947-445e-867e-f50019ff21eb" />

Let me know what you think!

Best,
Valentin

facontidavide:
That is a great idea, but I guess the devil is in the details.
The main problem is not the serial port, but the protocol (serialization).

PlotJuggler needs something that can be converted in a vector of key/values, where key is a string and value is generally a number

This means that an efficient serialization should avoid sending the key (names) in avery message.
This is easily done using a schema and something such as **Flatbuffers**.

Of course, if bandwidth is not a problem, you can use **messagepack**, that is schemaless.

If there is a "standard" way people do this serialization, I will be happy to discuss it.

Valle125:
The tools I've used so far usually supported one or more of the following approaches
- key-value pairs are sent, lines are separated by an of \n, \r\n or \r.
- comma separated values are sent, column names are known a prior and specified within tool settings.
- binary values are sent, column names and value format (byte length, endiannes etc. ) are known a prior and specified within tool settings.

For embedded systems with restricted resources it's especially simple to implement the binary or the comma separated values scheme.

Valle125:
Here is one [tool](https://github.com/hyOzd/serialplot) I've used so far. And some [screenshots ](https://hackaday.io/project/5334/gallery#8e494aa25753050e911ed29a3304184a).

The tool has limited options to display and process data, an area where plotjuggler excels.

ssys-zdavis:
I've been working toward a serial port streaming plugin for some time.

[...]

PR #1314, opened 2026-03-07
Feature: Serial Port

Valle125:
This pull request adds a serial port DataStreamer plugin (implements #1232).

Contrary to IP based communication, which delivers data (messages) packet wise, serial ports deliver data byte wise. This introduces the necessity to split data into messages before passing them to the protocol parser. The plugin solves this by introducing the MsgSplitter class. For each protocol a class must be derived from MsgSplitter and implement a specific splitting algorithm. Currently only the JSON protocol is supported. I'm planning to add support for more protocols in the future.

The picture below shows a preview of the DatastreamSerialPortDialog.
image


facontidavide:
thanks, this is more inline with what I wanted (delegate the serialization).

Let me have a look at it

facontidavide:
I will merge it as it is, but keep in mind that I will apply some minor  changes. thanks a lot for contributing

image

Issue #622, opened 2022-03-28
Extra metadata header appearing in topic tree, other issues stemming from this too


zm2022:
# Description:

I have successfully worked with Plotjuggler on an Ubuntu 18.04 system running ROS2 Foxy. I recently installed it on an Ubuntu 20.04 system, running ROS2 Galactic. I'm running into a new issue with it on this system. I believe this issue is also causing several other issues.

When I load a ROS2 bag file into Plotjuggler, the metadata header is showing up in the topic tree, as shown (the metadata header wasn't showing up in the tree on Ubuntu 18.04/ROS2 Foxy):

![image](https://user-images.githubusercontent.com/95399202/160307854-fcd0465e-17ce-4631-b72f-51aba76a74ee.png)

Two other issues that I believe are related: 
1. When I save a layout and try to reopen the ROS2 bag with the layout, I get the message "cannot find curves." This is because the topics can't be found as the metadata header isn't included in the topic's path in the layout file. In the layout file, I have to manually add "metadata" to the front of the topic path so Plotjuggler can find the topics and load the layouts.

As an example, if the layout included a topic reference like this: '/sample_topic/sample_topic_field1'
then in the layout file I need to update this reference to: 'metadata/sample_topic/sample_topic_field1'.

2. Also, when I try to edit curves on the plots, I get the message "cannot find curves," and the curve is not available in the editing menu, as shown:

![image](https://user-images.githubusercontent.com/95399202/160307875-1bccc2a4-7f95-4bae-9d87-0088e0ac0a80.png)

# System Info/Steps to reproduce:

SAMPLE ROS2 BAG HERE: 
[plotjuggler_metadata_issue_info.zip](https://github.com/facontidavide/PlotJuggler/files/8358854/plotjuggler_metadata_issue_info.zip)

- Previously on Ubuntu 18.04 and ROS2 Foxy (worked fine). Now on Ubuntu 20.04 and ROS2 Galactic (has the described issues)
- I'm attaching a sample bag of data and a sample message to use to access the data (I created a custom message with multiple message fields of different types, to mimic my custom messages as closely as possible). You'll probably need to updated the message path in the rosbag metadata file - I added a comment by the file path.
- NOTE: I was able to reproduce the metadata header issue and layout issue with this sample bag. However I was not able to reproduce the issue of being unable to edit curves. I'm hoping that this issue will be resolved through resolving the metadata issue, or can provide more info if needed.

[...]

PR #1320, opened 2026-03-11
Allow curve tree nodes to be both channels and parents

phfix:
### Summary

Fix a curve tree collision when a channel path is also a parent of other channels.

### Details

When / separators are used, names like input/a and input/a/ad_raw can collide in the tree representation.

The issue is order-dependent: if input/a/ad_raw is inserted first, the input/a node is created as a parent branch, and the later insertion of the real input/a channel is effectively hidden.

This change allows a tree node to represent both a real channel and a parent node, so the shorter path is no longer lost.

A small CSV sample was added to datasamples/ to reproduce the issue.

### Notes

This keeps the current drag behavior unchanged. If subtree dragging is added later, branch-vs-leaf behavior will likely need to be made more explicit in the UI.

Filtering on a node that is both a channel and a parent can still be somewhat ambiguous, but the current behavior is consistent with preserving both roles on the same row.

Possibly related to other path/tree representation issues such as #622, although the underlying cause here appears to be different.


facontidavide:
funny enough, I detected that yesterday. let me have a look at your proposed change

facontidavide:
image

PR #1322, opened 2026-03-16
Draft: Annotation feature preview

phfix:
This draft PR is shared as a preview of an annotation feature and the current implementation state.

The main goal is to make the feature available for early review and discussion. 
image

This PR is not being presented as merge-ready. The branch currently also contains some testing/CMake work and a few incidental fixes. I developed this work to solve a concrete problem, and I’m happy to share it in its current form for discussion and feedback. I may have time to continue refining it later, but I do not want to make promises at this stage about turning it into a merge-ready PR.

Current highlights:
- first implementation of the annotation feature
- support for managing annotation files directly in PlotJuggler
- example annotation CSV file included for trying the feature locally (datasamples\annotations_demo.csv)

Notes:
- this is a preview branch
- not substantially validated across build configurations
- especially the newer testing/CMake-related changes should be treated as provisional


facontidavide:
very interesting.

I am working on PlotJuggler 4. I should take this isea into account.
I will be back to you soon with some feedback

PR #1323, opened 2026-03-16
Add Python scripting, StringSeries support, and UI improvements to Custom Functions  

Alvvalencia:
- **Python engine:** Embedded Python 3.12 interpreter alongside the existing Lua engine.
  Users can write `calc(time, value, ...)` functions in Python with full traceback error
  reporting and line-number mapping to the editor.

- **StringSeries support:** Custom Functions now accept string timeseries as main sources
  or additional sources. A `MixedSource` abstraction handles dispatch between numeric
  (`PlotData`) and string (`StringSeries`) inputs in both engines.

- **Mixed-type additional sources:** Additional sources passed to `calc()` can be any
  combination of numeric and string series, passed to the scripting engine as native types.

- **Language persistence:** The selected scripting language (Lua/Python) is saved as part
  of the snippet and stored in the function library. Fully backward compatible — snippets
  without a language attribute default to Lua.

- **UI refactor and optimizations:** Improved Function Editor layout and language selector.
  Fixed white-rendering artifacts when switching scripting language.


maracuya-robotics:
Thanks for adding this functionality! I'm especially excited about the StringSeries support. A few observations I made while testing the pull request:

- Depending on whether there are errors or not, the preview plot will be shown or vanish. This causes the layout to jump a lot.
- The first timeseries item in the variables list is labeled `v1`, but only accessible as `value` in the scripts.
- The library is now hidden. If you just want to select a function, this is fine. But sometimes you want to copy parts of several functions or use them as a reference.
- Python support is a nice touch and it compiled without issues on my Arch Linux system.
- Splitting the features into multiple pull requests (for example StringSeries, Python, layout changes) or cleaning up the commits makes a review easier.

maracuya-robotics:
Apparently there are still some bugs.

- For example, the first try of dragging a timeseries into the list doesn't work.
- When a custom series is selected in addition to a regular series and the regular one is dragged, both series are added to the list.
- Selecting a timeseries and then clicking on "Create a new timeseries" does not automatically add the selected series to the input list.
- The layout still jumps around while writing functions because the preview plot is shown and hidden frequently.
- Now the preview plot takes more screen space without any way of changing its size.
- The function library popup is tiny and cannot be resized.
- I'm still missing the previous way of seeing scripts from the library together with the code editor for direct comparison and editing. It might be better to use a nonmodal dialog for the library.

Overall it would really help to clean up the commits and split the changed and new features into individual commits or even pull requests. I'm sure @facontidavide would appreciate it as well.

avalen2022:
Hi @maracuya-robotics , I’ve gone through everything and tested it on my side, and I’d like to clarify a couple of things:

I couldn’t reproduce the issue with the first drag attempt. It works correctly for me.
Please keep in mind that the timeseries list and the custom series list are independent. So selecting something in one does not automatically deselect items in the other.

maracuya-robotics:
Thanks, several of the issues are fixed now:

- Regular and custom timeseries added accidentally.
    → *fixed*
- Layout jumps while typing.
    → *error log and preview plot are now use the same space and are swapped*
- Preview plot takes larges space in layout.
    → *can be adjusted now*
- The function library popup is tiny and cannot be resized.
    → *It is now a regular OS client window (not drawn within the main PlotJuggler window) and can be resized as expected.*
- It is not possible to access the function library and main PlotJuggler window simultaneously.
    → *The popup is still a modal dialog and prevents access to the main PlotJuggler window, but for now this should be ok.*

Other issues are still present:

- Selecting a timeseries and then clicking on "Create a new timeseries" does not automatically add the selected series to the input list.
- The first try of dragging a timeseries into the list doesn't work. (See video for example)


https://github.com/user-attachments/assets/1781291e-b797-499c-8e25-70c06dbc89f5

maracuya-robotics:
Looks like this got merged into main in the meantime.

facontidavide:
merged

PR #1325, opened 2026-03-18
Added another common statistics metrics, rms, mav, stddev, peak to peak

Lanove:
This PR adds additional statistical metrics to PlotJuggler, including:
- Mean
- Standard Deviation
- RMS (Root Mean Square)
- Peak-to-Peak
- MAV (Mean Absolute Value)

While working with data visualization in PlotJuggler, I found myself needing more built-in statistical indicators directly within the plotting interface. These metrics are commonly used for signal analysis and quick diagnostics, and having them readily available improves workflow efficiency without requiring external tools.

Oh and another bugfix
Average interval is now computed using sample intervals instead of sample count:
Average Interval = (end_time - start_time) / (count - 1)

This is mathematically correct because N samples contain N-1 intervals.
Example: 3 samples at 1-second spacing span 2 seconds, so average interval is 2 / (3 - 1) = 1 second.

Screenshot_20260318_122056

### DIFF
--- a/plotjuggler_app/statistics_dialog.cpp
+++ b/plotjuggler_app/statistics_dialog.cpp
@@ -7,6 +7,7 @@
 #include "statistics_dialog.h"
 #include "ui_statistics_dialog.h"
 #include <QTableWidgetItem>
+#include <cmath>
 #include "qwt_text.h"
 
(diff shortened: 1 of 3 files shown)


facontidavide:
thanks

PR #1331, opened 2026-04-07
Fix: Complete Timeseries list synchronization and custom series separation

Qubi0-0:
I encountered a problem when using layouts: the selected topics did not appear in the Timeseries List section. As a result, I could reuse the same topic multiple times, for example in another place. This change allows the same topics to be used multiple times, even after they have already been used by a layout.

## Summary

Improves how PlotJuggler restores and displays series from saved layouts/configurations.

- Missing topics from a layout now appear in the Timeseries list immediately as placeholders, even before live data arrives.  

## Problem

During layout restoration:
1. Topics referenced in the layout were sometimes missing from the Timeseries list. There was no option to use same topic if it was already used by loaded Layout.

## Implementation
 
 - Placeholder series are created for missing topics 
during layout load (numeric, XY endpoints, and string
 curves).
 - A shared helper collects all curve names from the 
layout XML, reused by both the placeholder block and 
the existing `checkAllCurvesFromLayout()`.
 - `CurveListPanel::addCurve()` skips series 
registered as custom (`_transform_functions`).  
 - `CurveListPanel::addCustom()` removes placeholders
 from the main list if needed.

## Result

- Layout topics now consistently appear in the Timeseries list.


facontidavide:
thanks, this seems to be very useful

facontidavide (changes requested):
Thanks for the contribution! The underlying idea — eagerly creating placeholders for layout topics and keeping transform/custom series out of the main Timeseries list — is a genuine UX improvement. A few blocking issues need to be addressed before this can be merged.

## 🔴 Blocking: Code does not compile

`mainwindow.cpp` has two stray closing braces around the new block. After the 4 added lines, the original `}` `}` closing `if (dataStreamers()...)` and `if (msgBox.clickedButton() == yes)` are still there, so the `else` branch at what is now line 2186 has no matching `if` and the whole function's brace nesting is broken. This is what the 12 failing CI jobs are reporting.

Fix (remove the two duplicated closing braces):

```diff
         _curvelist_widget->refreshColumns();
-          }
-        }

         startStreamingPlugin(streamer_name);
       }
       else
```

Please run `pre-commit run -a` locally (which invokes `clang-format`) — it would have caught the mismatched indentation on those two lines, and it's a required check on this repo.

## 🟠 PR description doesn't match the diff

The description states:

> `CurveListPanel::addCustom()` removes placeholders from the main list if needed.

…but `addCustom()` is unchanged in the diff. So if a topic is first added to the main list and later becomes a custom function (e.g. a Lua reactive function producing a curve with the same name), it will remain visible in **both** lists. Either implement this part of the fix, or update the description to reflect what was actually changed.

## 🟠 Placeholder creation is incomplete and only runs on one narrow path

The new block only runs when:
1. The layout contains a `previouslyLoaded_Streamer` entry, **and**
2. The user clicks "Yes" on the "Start Streaming?" dialog.

For layouts loaded without a streamer (or when the user clicks "No"), behaviour is unchanged — the old `checkAllCurvesFromLayout` dialog still fires.

Additionally, the new block uses `readAllCurvesFromXML()`, which collects only regular `curve` elements' `name` attribute. It does **not** collect:
- XY plot endpoints (`curve_x` / `curve_y`)
- String curves

Compare with `checkAllCurvesFromLayout()` (mainwindow.cpp:1041) which handles all three cases. As a result, XY layouts won't get placeholders and strings won't either — the issue you described will still reproduce for those layouts.

**Suggestion:** extract the placeholder-creation body into a single helper method (e.g. `createLayoutPlaceholders(const QDomElement& root)`) that mirrors what `checkAllCurvesFromLayout` already does (XY + numeric + string) and call it from the single place that's most appropriate. That also avoids duplicating the logic.

## 🟢 What's sound

- The `_transforms_map.count(plot_name) > 0` guard in `addCurve()` is correct and safe. I audited all 7 callers (mainwindow.cpp lines 820, 1106, 1359, 1859, 2177, 2543, 2556) and none depend on `addCurve` succeeding for a name that is simultaneously a registered transform.
- `_mapped_plot_data.addNumeric` and `_tree_view_items.count` both make the new `addCurve` call idempotent, so subsequent `MoveData`-driven calls during streaming remain no-ops. No regression there.

## Suggested checklist before re-requesting review

- [ ] Remove the two duplicate `}` in `mainwindow.cpp` and confirm `cmake --build` succeeds locally
- [ ] Run `pre-commit run -a` (enforces clang-format)
- [ ] Align the PR description with what is implemented, or implement the `addCustom` cleanup described
- [ ] Decide whether placeholder creation should also cover XY and string curves, ideally via a shared helper reused by `checkAllCurvesFromLayout`

Qubi0-0:
I made changes as suggested in the comment. Also during code Review other issue has came up in 

```cpp
  for (auto& curve_name : curves)
  {
    if (_mapped_plot_data.numeric.count(curve_name) == 0)
    {
      missing_curves.push_back(curve_name);
    }
    if (_mapped_plot_data.strings.count(curve_name) == 0)
    {
      missing_curves.push_back(curve_name);
    }
  }
```
To:
```cpp
  for (auto& curve_name : curves)
  {
    if (_mapped_plot_data.numeric.count(curve_name) == 0 &&
        _mapped_plot_data.strings.count(curve_name) == 0)
    {
      missing_curves.push_back(curve_name);
    }
  }
```



 **Root cause:** Two independent `if` checks (`numeric` and `strings`) caused every 
curve to be
 pushed into `missing_curves` twice — once for each map it was absent from. A 
placeholder created
 via `addNumeric()` would pass the numeric check but still fail the strings check, so 
the dialog
 fired regardless.
 
 **Why it matters for this MR:** The new placeholder block in `loadLayoutFromFile()` 
adds curves
 via `addNumeric()`. Without this fix, `checkAllCurvesFromLayout()` (called later in 
`xmlLoadState()`)
 would still show the warning for those same curves, breaking the intended UX flow.
 
 **Fix:** Combined into a single `&&` condition — a curve is only considered missing 
when absent
 from both `numeric` and `strings`.

P.S.
Terribly sorry for commit mess. I did some unnecessary quick decisions. With squashing the commits it shouldn`t make any diffrence though.

PR #1332, opened 2026-04-08
[Draft] Feature: Add alternate copy paste with raw curve values.

DavidJKeizer:
## Current Behavior

In the curve list panel, the current copy shortcut simply copies the source path for the series. This can be used to search for a series in the list.

Pressing (**Ctrl + C**) on *gps_quality_poor* copies the following to the clipboard:

DefaultCopyPasteUsage

```estimator_event_flags.02/gps_quality_poor```

## Present Restrictions

While functional, there are times where copy the series name *and* its value to the clipboard is preferable. Consider the case where a latitude and longitude pair must be cross-referenced with a map. Presently, there is no convenient method available to do this. In the example case of GPS coordinates, the number of digits of prevision that plot juggler shows may also be insufficient. 

LatLonCopyUsecase

## New Feature
This PR adds a new keyboard shortcut (**Alt + C**) which appends *all* selected series and their complete raw values to the clipboard.
This allows users to see the raw values, and manipulate them as needed for use in other software. 

A new shortcut was chosen to be added instead of modifying the existing copy behavior to prevent breaking the current copy -> search behavior. 

In the above example (**Alt + C**) will copy the following to the clipboard:
```
vehicle_global_position/lat: 30.72258130993994
vehicle_global_position/lon: 103.90223135162262
```

### Testing
The PR has been built and tested on Windows 11 with vcpkg, as well as with Ubuntu 24.04 LTS in WSL using the docker container.

### Change Log Entry
* Added ALT + C shortcut to copy selected series and their raw values.


DavidJKeizer:
CI/CD is failing, so I have marked this as draft until I can look at it again. I also want to add a setting to configure default copy/paste behavior.

Issue #750, opened 2022-09-21
[Feature Request] Display the filename that is open


justas-:
Thanks for contributing to PlotJuggler. You are great!

# Problem description

I would like to see the name of the ULog file I have open somewhere in the UI.

# Steps to reproduce (important)

1. Open any ULog file. 
2. Look at the UI.
3. Notice that you can't say what ULog file is open.

pbouffard:
I'd like to see the same feature. In the meantime a workaround is to check the topmost item in the drop-down next to the data load button:

<img width="344" alt="image" src="https://user-images.githubusercontent.com/231346/197232010-bb0d169a-962a-4037-9b42-09a229311627.png">

I might try to implement this if I get a chance. One complication would be that there can be more than one file loaded at a given time, and also what to display in cases where streaming is being used (which I don't have experience with).

vineet131:
@pbouffard Did you get a chance to try this? This would be a really useful change. Because I have multiple tabs from multiple files opened, and its a bit complicated to keep track of.
> One complication would be that there can be more than one file loaded at a given time
How about just appending the filenames in that case? And expanding on hovering over the title/tab
> what to display in cases where streaming is being used 
I haven't tried the streaming part of this. But if the user has to enter an `IP:port` maybe that could be displayed? Or you could just leave it blank/default for the timebeing.

facontidavide:
As I'm moving toward making it easier to open multiple files, this is actually more tricky that it seems

PR #1333, opened 2026-04-13
[Draft] QOL: Show currently loaded file(s) in the file label.

DavidJKeizer:
## Current Behavior

When loading log files with parameters the file name is displayed in the parameter dialog. 

ulog_parameters_dialog

In the event that multiple files are loaded in at the same time, then each file is given its own root path to distinguish them. 

time_series_directories

## Present Restrictions

CSV files do not have the parameter dialog, and as such there is no indicator to indicate which file is currently open when only a single file is loaded.

For files with many series available, the root paths are not necessarily visible. Consider these PX4 ULog files for an example, where there are multiple files loaded, but no immediately visible labels to indicate which ones.

As a result, when multiple instances of PlotJuggler is running at the same time it is unclear which instance has which file(s) loaded.

time_series_ambiguous_series

The conventional way to do this in many applications is too rename the main window. However, to my understanding this would potentially interfere with skins, and may not be as practical when several files are open in the same instance.

## New QOL Feature

This PR repurposes the "File" label to show what file(s) are presently loaded in the current instance of PlotJuggler. In the event that multiple files are loaded, or the file path is too long, the label truncates the text. To cover this edge case, a tool-tip has been added to show the full list of all loaded files.

FilesOpenLabelWithTooltip

### Testing
The PR has been built and tested on Windows 11 with vcpkg, as well as with Ubuntu 24.04 LTS in WSL using the docker container.

### Change Log Entry
* QOL: Show currently open file path(s) in *File* header.

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -1422,6 +1422,14 @@ bool MainWindow::loadDataFromFiles(QStringList filenames)
 
   ui->buttonReloadData->setEnabled(!loaded_filenames.empty());
 
+  QList<QFileInfo> loaded_fileinfos{};
+  for (const auto& file_info : loaded_filenames)
+  {
+    loaded_fileinfos.push_back(QFileInfo(file_info));
+  }
+  updateCurrentLoadedFilesLabel(loaded_fileinfos);
+
+
   if (loaded_filenames.size() > 0)
   {
     updateRecentDataMenu(loaded_filenames);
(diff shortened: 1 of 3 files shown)


facontidavide:
fix formatting with pre-commit.

Have you considered all the cases here:

image


Also, what happens when you click "clear"? is the list of loaded files removed

DavidJKeizer:
PrefixSupport

- Rebased onto master again.
- Fixed delete all not clearing the label.
- Added prefixes to both the label and tool-tip when provided.

Something funky is going on with the merge series behaviour, so I have marked this PR as a draft until I can have a look at that again.

facontidavide:
Thanks for starting this. PJ4 removed the old file-label widget entirely — the left panel was redesigned, and loaded datasets now appear as named tracks on the Source Timeline — so this specific patch has no target. The remaining wish (identifying the open file, e.g. in the window title) is still open as #750, and PR #1377 proposes it; that is the better place to continue.

PR #1351, opened 2026-04-24
CI: install missing jazzy typesupport implementations

GNERSIS:
ros-jazzy-ros-base does not guarantee that rosidl typesupport implementations are pulled in, causing CMake to fail with "No 'rosidl_typesupport_c' found" when configuring rclcpp. Explicitly install introspection-based typesupports as a fix.

Mirrors the earlier rolling fix in d00d8a6f.


facontidavide:
Can't you add this to package.xml instead?

PR #1353, opened 2026-04-27
Mosaico toolbox retain batches

GNERSIS:


facontidavide:
meged with some changes to fix CI

PR #1369, opened 2026-05-17
fix(plot): restore Y range envelope for constant-valued series

### LINKED ISSUE #1080: Vertical zoom stuck on max 0.1
Hello everyone,

I just downloaded plotjuggler for ROS2 Jazzy and I cannot get the vertical zoom to zoom out further than 0.1 when data is loaded. Before adding data to the plot, it can zoom until 1.

I just installed with sudo apt install ros-jazzy-plotjuggler-ros and the issue immediately appeared when I loaded my ros2bag.

Please let me know if there is any other info I can provide.

Thank you and regards

dakejahl:
## Summary

A curve whose Y values are all the same constant currently does not render — the plot looks empty after dropping the series in.

## Problem

`PlotWidgetBase::getVisualizationRangeY` computes `margin = (top - bottom) * 0.025`. When every sample equals some constant `K`, `top == bottom == K`, so the margin is `0` and the returned Y range collapses to `{K, K}`. This zero-height range is then handed to `setAxisScale(yLeft, K, K)` (via `updateMaximumZoomArea` / `resetZoom`), and the curve becomes invisible.

The regression was introduced in e4ee6288 ("Fix #1080: vertical zoom no longer stuck on max 0.1"), which removed the previous `margin = 0.1` fallback. That commit fixed the empty-data case but didn't account for the constant-data case.

## Solution

When `top - bottom` is below epsilon, fall back to an envelope proportional to the value (`5% of |K|`), or `+/-1.0` when the constant itself is approximately zero. The empty-data branch (`bottom > top`) is untouched and still defaults to `+/-1.0`, so the #1080 behavior is preserved.

### DIFF
--- a/plotjuggler_base/src/plotwidget_base.cpp
+++ b/plotjuggler_base/src/plotwidget_base.cpp
@@ -271,7 +271,25 @@ Range PlotWidgetBase::getVisualizationRangeY(Range range_X) const
     top = 1.0;
   }
 
-  double margin = (top - bottom) * 0.025;
+  double margin;
+  if (top - bottom > std::numeric_limits<double>::epsilon())
+  {
+    margin = (top - bottom) * 0.025;
+  }
+  else
+  {
+    // Constant series: top == bottom would collapse the Y axis to zero height
+    // and hide the curve. Give it a visible envelope proportional to the value,
+    // falling back to 1.0 when the value itself is ~0.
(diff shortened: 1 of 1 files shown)


facontidavide:
PJ4 already guards this: a constant-valued series gets a non-zero fallback range (+/-0.1) rather than a zero-height one. Its policy differs from the proportional envelope proposed here, but the crash-y zero-height case is handled.

PR #1374, opened 2026-05-21
feat(plugins): auto-discover ROS plugin packages

dakejahl:
## Summary

Auto-discover PlotJuggler plugins shipped by sibling colcon/catkin packages by walking the ROS environment prefix path and globbing `<prefix>/lib/plotjuggler_*/` in each. Removes the hardcoded `plotjuggler_ros` lookup.

## Problem

`loadAllPlugins` only knows about one plugin package by name. The AMENT branch calls `ament_index_cpp::get_package_prefix("plotjuggler_ros")` and the CATKIN branch appends `lib/plotjuggler_ros` to each `CMAKE_PREFIX_PATH` entry. Any other plugin package in the workspace — `plotjuggler_apbin_plugins`, any future one — is invisible. Users have to add the folder manually under Preferences → Plugins, which isn't discoverable. Loading e.g. an ArduPilot `.bin` file then finds zero compatible loaders and falls through to a "Select the loader" dialog.

## Solution

Walk the colon-separated `AMENT_PREFIX_PATH` (or `CMAKE_PREFIX_PATH` for ROS 1) and enumerate `<prefix>/lib/plotjuggler_*/` subdirectories. Sibling plugin packages already follow the convention of installing to `lib/<package_name>/` with names prefixed `plotjuggler_*` (this is what `plotjuggler_ros` itself does), so no opt-in is required from existing plugin repos.

Drops the legacy "Missing package [plotjuggler-ros]" warning that was added for the 2.x → 3.x upgrade path and no longer applies — a fresh ament install with only non-ros plugins would otherwise see it spuriously.

### DIFF
--- a/plotjuggler_app/mainwindow.cpp
+++ b/plotjuggler_app/mainwindow.cpp
@@ -64,13 +64,6 @@
 #include "cheatsheet/cheatsheet_dialog.h"
 #include "colormap_editor.h"
 
-#ifdef COMPILED_WITH_CATKIN
-
-#endif
-#ifdef COMPILED_WITH_AMENT
-#include <ament_index_cpp/get_package_prefix.hpp>
-#include <ament_index_cpp/get_package_share_directory.hpp>
-#endif
 
 // Serialize a QDomDocument to XML with attributes in sorted order.
 // Qt's QDomDocument::toString() uses a hash map for attributes, producing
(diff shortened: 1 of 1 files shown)


facontidavide:
This discovers legacy catkin/ament Qt plugins. PJ4 loads SDK extensions through marketplace-managed scan paths instead, so PJ3-era plugin packages cannot be discovered and loaded unchanged. The underlying wish — not having to point at a plugin dir by hand — is handled by the extensions marketplace.

PR #1378, opened 2026-06-02
fix: keep custom series equation numbering when editing

tstienstra:
When loading a layout from older plotjuggler versions with the linked resource as `value` and the additional resources starting from `v1`, I ran into the problem that it does load the layout correctly. However, when editing an equation the sources are renamed causing invalid output.


facontidavide:
This preserves the legacy `value`/`v1` numbering in PJ3's function editor. PJ4 replaced that editor and the processor contract behind it, and does not aim for full PJ3-layout compatibility, so the patch has no counterpart there.

PR #1382, opened 2026-06-10
fix nix package

gvz:
this fixes the nix build


facontidavide:
Same as #1411 — this fixes the old Nix package, and PJ4 has no flake for it to target. A Nix package built against PJ4 would be welcome as a new PR.

PR #1397, opened 2026-07-07
Fix CDR Alignment for compatibility with other DDS Vendors

jamgotchian:
Summary:
  - Fixes decoding of MCAP files from RTI Connext DDS Micro 7.3
  - Root cause: RTI aligns from byte 0, FastDDS aligns from byte 4 (both valid per ambiguous spec)
  - Solution: Configurable alignment in nanocdr, enabled for OMGIDL encoding
  - Backwards compatible: ROS2 users unaffected (different encoding type)
  - Low risk: Only affects OMGIDL-encoded files (currently broken for RTI)

## Problem

PlotJuggler was unable to decode MCAP files recorded with RTI Connext DDS Micro 7.3. When loading these files:
- Boolean fields decoded correctly
- Double and other multi-byte fields showed garbage values
- Error: "Decode: not enough data to decode"

### Root Cause

RTI DDS Micro and eProsima FastDDS interpret CDR (Common Data Representation) alignment differently:

- **RTI DDS Micro**: Calculates alignment from byte 0 (message start, including the 4-byte CDR header)
- **FastDDS/ROS2**: Calculates alignment from byte 4 (after the CDR header)

The OMG CDR specification is ambiguous on this point, leading to divergent but valid implementations.

PlotJuggler's nanocdr decoder (based on FastCDR) always used byte 4 alignment, causing it to read data at incorrect offsets when decoding RTI-generated messages.

**Example**: In a message with 48 booleans followed by a double:
- RTI places the double at byte 56 (48 booleans + 4 padding bytes to align to 8-byte boundary from byte 0)
- PlotJuggler expected the double at byte 52 (48 booleans, already 4-byte aligned from byte 4)
- This 4-byte mismatch caused decoding failures

## Solution

Added configurable alignment origin to the nanocdr CDR decoder:

1. **nanocdr.hpp**: Added `align_from_message_start` parameter (default: `false` for backward compatibility)
   - `false` → aligns from byte 4 (FastDDS/ROS2 behavior)
   - `true` → aligns from byte 0 (RTI DDS Micro behavior)

2. **deserializer.hpp/cpp**: Added `setAlignFromMessageStart()` method to configure the decoder

3. **omgidl_parser.h**: Enable byte 0 alignment for all OMGIDL-encoded messages

4. **ros2_parser.h**: Unchanged - ROS2 messages continue using byte 4 alignment (default)

### Additional Notes

- **Encoding type discrimination**: MCAP files distinguish between "omgidl" (generic DDS) and "ros2msg" (ROS2-specific) encodings
- **No vendor detection needed**: The encoding type is the correct discriminator, not vendor-specific markers
- **Standards conformance**: RTI likely follows the OMG standard more strictly; FastDDS is the ROS2-specific variant
- **Backward compatible**: Default behavior unchanged, only affects OMGIDL-encoded files

### Core CDR Decoder
- `plotjuggler_plugins/ParserROS/rosx_introspection/include/rosx_introspection/contrib/nanocdr.hpp`
  - Added `bool align_from_message_start = false` parameter to `Decoder` constructor
  - Modified constructor to set `origin_` based on alignment flag
  - Added explanatory comments about the two alignment approaches

### Deserializer Interface
- `plotjuggler_plugins/ParserROS/rosx_introspection/include/rosx_introspection/deserializer.hpp`
  - Added `setAlignFromMessageStart(bool enable)` method
  - Added `bool _align_from_start = false;` member variable

- `plotjuggler_plugins/ParserROS/rosx_introspection/src/deserializer.cpp`
  - Pass `_align_from_start` flag to nanocdr `Decoder` constructor in `reset()` method

### Parser Factory
- `plotjuggler_plugins/ParserROS/omgidl_parser.h`
  - Enable byte 0 alignment for all OMGIDL-encoded messages
  - Added comments explaining the difference from FastDDS/ROS2

## Risk Assessment

**Only potential issue**: If someone incorrectly labeled FastDDS data as "omgidl" encoding in MCAP files, this would break their workflow. However, that would be a data labeling error, not a problem with this fix.

### CDR Message Structure
```
Byte 0:   0x00 (always)
Byte 1:   Encoding flags (includes endianness)
Byte 2:   Options
Byte 3:   Reserved
Byte 4+:  Message payload
```

## References

- RTI Connext DDS Micro: https://www.rti.com/products/connext-dds-micro
- eProsima FastDDS: https://www.eprosima.com/index.php/products-all/eprosima-fast-dds
- OMG DDS Specification: https://www.omg.org/spec/DDS/
- MCAP Format: https://mcap.dev/

### DIFF
--- a/plotjuggler_base/include/PlotJuggler/messageparser_base.h
+++ b/plotjuggler_base/include/PlotJuggler/messageparser_base.h
@@ -115,6 +115,12 @@ public:
     _use_embedded_timestamp = enable;
   }
 
+  // Configure CDR alignment for OMGIDL parsers (default no-op for other parsers)
+  virtual void setUseLegacyCdrAlignment(bool /*enable*/)
+  {
+    // Default implementation: no-op (only OMGIDL/ROS2 parsers support this)
+  }
+
 protected:
   PlotDataMapRef& _plot_data;
   std::string _topic_name;
--- a/plotjuggler_plugins/DataLoadMCAP/dataload_mcap.cpp
+++ b/plotjuggler_plugins/DataLoadMCAP/dataload_mcap.cpp
@@ -292,6 +292,7 @@ bool DataLoadMCAP::xmlSaveState(QDomDocument& doc, QDomElement& parent_element) 
   elem.setAttribute("use_mcap_log_time", int(params.use_mcap_log_time));
   elem.setAttribute("clamp_large_arrays", int(params.clamp_large_arrays));
   elem.setAttribute("max_array_size", params.max_array_size);
+  elem.setAttribute("use_legacy_cdr_alignment", int(params.use_legacy_cdr_alignment));
   elem.setAttribute("selected_topics", params.selected_topics.join(';'));
 
   parent_element.appendChild(elem);
@@ -311,6 +312,7 @@ bool DataLoadMCAP::xmlLoadState(const QDomElement& parent_element)
   params.use_mcap_log_time = bool(elem.attribute("use_mcap_log_time").toInt());
   params.clamp_large_arrays = bool(elem.attribute("clamp_large_arrays").toInt());
   params.max_array_size = elem.attribute("max_array_size").toInt();
+  params.use_legacy_cdr_alignment = bool(elem.attribute("use_legacy_cdr_alignment").toInt());
   params.selected_topics = elem.attribute("selected_topics").split(';');
   _dialog_parameters = params;
   return true;
@@ -562,6 +564,15 @@ bool DataLoadMCAP::readDataFromFile(FileLoadInfo* info, PlotDataMapRef& plot_dat
     {
       auto& parser_factory = it->second;
       auto parser = parser_factory->createParser(topic_name, schema->name, definition, plot_data);
+
+      // Configure legacy CDR alignment for OMGIDL parsers if requested
+      if (_dialog_parameters->use_legacy_cdr_alignment)
+      {
+        if (channel_encoding == "omgidl" || schema_encoding == "omgidl")
+        {
+          parser->setUseLegacyCdrAlignment(true);
+        }
+      }
 
       parsers_by_channel.insert({ channel_ptr->id, parser });
     }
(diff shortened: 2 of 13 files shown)


facontidavide:
Thanks @jamgotchian, great root-cause analysis — this is a real gap, and files like yours are simply undecodable today.

Before merging I'd like to discuss the gating. The `omgidl` schema encoding tells us the schema *language*, not which vendor serialized the payload. Any existing `omgidl`-encoded MCAPs written by FastCDR-style serializers (byte-4 alignment) decode fine today and would break with an unconditional `setAlignFromMessageStart(true)`. And as you observed yourself with the doubles, misalignment usually doesn't throw — it just produces garbage values — so affected users wouldn't get an error, just silently wrong plots.

Two thoughts on that:

1. If I read DDS-XTypes correctly, XCDR2 defines the alignment origin unambiguously as the first byte *after* the encapsulation header. So the byte-0 origin should arguably apply only when the header says classic CDR / XCDRv1, never for XCDRv2 — right now the flag flips both.
2. Since reliable auto-detection isn't possible (the encapsulation identifiers don't discriminate vendors, and wrong alignment often decodes "successfully"), the escape hatch probably has to be a user-visible option — e.g. a QSettings-backed checkbox next to the existing `Preferences::truncation_check`, with the default being whatever we believe is the most common producer of `omgidl`-encoded MCAPs. Do you know of tools other than RTI that write `omgidl`-encoded MCAPs? That would settle the default.

Two practical requests as well:

- Could you add a unit test with a small hand-crafted CDR buffer? Your 48-booleans-followed-by-a-double example would be perfect to lock the behavior in (and a tiny sample MCAP from DDS Micro for the test data would be even better).
- The `rosx_introspection` folder here is synced from https://github.com/facontidavide/rosx_introspection, which has the same hard-coded `+4` origin — the change needs to land there too, otherwise the next sync will revert it. I can take care of that part once we settle the API here.

jamgotchian:
Apologies for the delay! I was finishing up some other projects and didn't have time to revisit this. 

I agree with you regarding XCDR2 alignment origin - the payload aligned relative to the first byte after the header. More tools and vendors are making XCDR2 the default, so I would push for the existing implementation (FastDDS format) to remain the default for PlotJuggler as well. Including my changes as a user-selectable option gives users tied to legacy CDR to benefit from PJ, too. 

I haven't crossed paths with many tools which produce mcap files with the `omgidl` encoding. In my case, I chose to use mcap as an onboard data logging format since it works great with multi-channel data. It just so happened that RTI was my DDS provider for the same project. 

I added the user-selectable option for the alignment origin and a small unit test. Let me know if there's any rework needed. And thanks for the feedback!

PR #1402, opened 2026-07-19
Link Qwt against Qt XML

Tobias-Fischer:
This PR is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: RoboStack `patch/ros-rolling-plotjuggler.osx.patch`, authored by Daisuke Nishimatsu.

Best-guess rationale: the vendored Qwt target uses Qt XML APIs through its Qt integration. Linking `Qt5::Xml` explicitly makes the dependency available for CMake toolchains that require complete target link interfaces.

Related RoboStack upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16


Tobias-Fischer:
FYI @wep21 and @traversaro, this upstream PR comes from the RoboStack patch upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16.

Origin patch: RoboStack `patch/ros-rolling-plotjuggler.osx.patch`, authored by Daisuke Nishimatsu.

traversaro:
Thanks @Tobias-Fischer ! The windows CI failures seems unrelated, and it is probably related to the upgrade of `windows-latest` GitHub Actions image to VS2026.

facontidavide:
the good news is that we probably don't need this in PlotJuggler 4. I **precisely** merged today a change removing the vendored QWT and using instead Pixi.

The next question is if we want PlotJuggler 3 to support Pixi too

PR #1407, opened 2026-07-31
Update README.md

shubham-shahh:
just some minor stuff


facontidavide:
Good catch on the typos. The PJ4 README on `main-4.x` is a different file and does not contain either of them, so there is nothing left to apply.

PR #1410, opened 2026-08-07
Fix Nix flake build and Qt runtime

mevatron:
## Summary

- update the Nix package version to match PlotJuggler 3.17.2
- remove stale Fast-CDR and MCAP overrides left over from the 3.10.11 source layout
- use Nix-provided shared LZ4 and Zstd libraries
- fetch the required Wasmer archive through Nix for sandboxed builds
- restore Qt application wrapping so Wayland/XCB platform plugins are discoverable at runtime

## Root cause

The flake still patched paths and dependencies from PlotJuggler 3.10.11. Fast-CDR support was later removed, causing `postPatch` to fail while deleting a file that no longer exists. After that was corrected, the newer source also required updated LZ4, Zstd, and Wasmer handling. In addition, `dontWrapQtApps = true` left the executable without Qt platform plugin paths.

## Validation

- `nix build .#plotjuggler --no-link`
- `QT_QPA_PLATFORM=offscreen nix run .#plotjuggler -- --help`
- `git diff --check`

The package builds successfully and the Qt-initializing help command exits successfully.


mevatron:
Superseded by #1411 after renaming the fork branch to `pr/fix-nix-flake-build`.

PR #1416, opened 2026-08-20
Fix ZMQ timestamp bug

classy-crustacean:
There was a bug where a ZMQ stream could not accept a timestamp as it is supposed to. If you tried to include a timestamp in a ZMQ packet, PlotJuggler would not even show the payload.

This was because in the receiveLoop() function, the 'msg' variable held a reference to recv_msg. Then, if a timestamp was included in the packet, recv_msg would be rebuilt, affecting msg.

I fixed this by creating a new zmq::message_t if there is a timestamp, and using that to grab the timestamp and extinguish remaining parts of the message. This leaves 'msg' unaffected. I moved the extinguishing code to the inside of the if statement that checks for the timestamp, because there is no way that it would run if the timestamp was not found, and it constrains the scope of timestamp_extra_msg.
I also renamed recv_msg to topic_payload_msg to make it more clear what its purpose is.

This issue can be replicated by running the following python code, and then trying to subscribe to the stream with PlotJuggler:
```
import zmq
from time import sleep, time
import msgpack
from random import randrange

cxt = zmq.Context()
s = cxt.socket(zmq.PUB)
s.bind("tcp://*:5555")
while True:
    timestamp = str(time()).encode('utf-8')
    print(timestamp)
    s.send_multipart([b'test', msgpack.packb({'signal': randrange(-100, 100)}), timestamp])
    sleep(0.1)
```


facontidavide:
PJ4's ZMQ subscriber already keeps the payload and the optional timestamp in separate messages (`ts_msg`), so the aliasing this fixes cannot occur there. Closing as superseded rather than unwanted — the bug you found was real on 3.x.

